PR #8, opened 2017-08-07
Integration with changes from 0.48-stable react-native branch

MaxRis:
Because react-native package module was moved into separate node package and git repo, to run application it's required manually edit {clone_root}/node_modules/metro-bundler/src/defaults.js by adding ubuntu platform into the list. Patch file is attached.
 [add-ubuntu-platform.patch.tar.gz](https://github.com/status-im/react-native-linux/files/1206057/add-ubuntu-platform.patch.tar.gz)


vkjr:
@MaxRis, can we move this to "closed"?

MaxRis:
@vkjr it seems that this issue is bound to parent issue. I think that we can close parent

PR #14, opened 2017-08-14
fixes #12 Added basic implementation of UIManager::setChildren method.

### LINKED ISSUE #12: Fix missing setChildren method of UIManager
The error can be seen by running TicTacToe example and clicking on the app window (or by resizing the window manually).

narek-h:
(fixes the broken TicTacToe example)

Test:
 With this change the UI is properly displayed, but click handling of items is stil broken (another missing thing).
 To temporary fix that for testing you can  comment out JSTimersExecution usage from  ReactQt/runtime/src/reacttiming.cpp file.

### DIFF
--- a/ReactQt/runtime/src/reactuimanager.cpp
+++ b/ReactQt/runtime/src/reactuimanager.cpp
@@ -126,12 +126,12 @@ void ReactUIManager::setChildren
   const QList<int>& childrenTags
 )
 {
-  //This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
-  QList<int> newIndices;
+  //TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
+  QList<int> indices;
   foreach(int i, childrenTags) {
-     newIndices.append(i + 1);
+    indices.append(1); //same Z order for all items
   }
-  manageChildren(containerReactTag, QList<int>(), QList<int>(), childrenTags, newIndices,  QList<int>());
+  manageChildren(containerReactTag, QList<int>(), QList<int>(), childrenTags, indices,  QList<int>());
 }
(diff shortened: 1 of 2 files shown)


ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -120,6 +120,20 @@ void ReactUIManager::updateView
   m_bridge->visualParent()->polish();
 }
 
+void ReactUIManager::setChildren
+(
+  int containerReactTag,
+  const QList<int>& childrenTags
+)
+{
+  //This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
```
MaxRis: @narek-h  It might be good to use comments with TODO keyword to revisit in the future.
For example, `// TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting`

ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -120,6 +120,20 @@ void ReactUIManager::updateView
...
+  QList<int> newIndices;
+  foreach(int i, childrenTags) {
+     newIndices.append(i + 1);
```
MaxRis: @narek-h it seems spacing formatting differs here from rest of the project, 2 spaces indents should be used

MaxRis: @narek-h , since `newIndices` are used in `manageChildren` to set Z-order on displayed `QQuickItem`, it might be not correct to increment Z-order for every child. If to take TicTacToe app, as example, displayed qml items cells most of all should have the same Z-order value.

MaxRis:
@narek-h thank you for the changes

PR #32, opened 2017-08-17
Bug/redbox broken

vkjr:
### Description
RedBox was broken in current implementation. In case of error it should show error message and stack trace

### Changes
1. Redbox qml code changed, Ubuntu components dependencies remove
2. Default z-order for created components changed from 1 to 0. Otherwise redbox can't be show since it creates with default z-order wich is 0
3. Test ImageApp added for future fixes on <Image> component. (not related to inbox, but related to my next tasks and was added to branch before redbox fixes) 

### How redbox looks now:
screen shot 2017-08-17 at 5 36 44 pm


ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -129,7 +129,7 @@ void ReactUIManager::setChildren
   //TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
   QList<int> indices;
   foreach(int i, childrenTags) {
-    indices.append(1); //same Z order for all items
+    indices.append(0); //0 is a default z-order in qml
```
MaxRis: @narek-h , could you please say, is that breaks TicTacToe cells rendering in sample app?

vkjr: @MaxRis, @narek-h, I was testing on TicTacToe. It doesn't break it.

narek-h: @MaxRis , @vkjr.  When I was setting 0 Z order, the app title was drawn at the bottom (instead of top). Of course, that issue is not caused by Z order, just the same indices were used as indices in layout. 
will double check that.

narek-h: but, in the context of this issue, this is a correct fix. So, let's merge it and in case of incorrect order of items, fix that in a separate patch.

vkjr: @narek-h, you are right, title is at the bottom now. Created issue #33.

vkjr:
### Changes

1. Redbox qml code changed, Ubuntu components dependencies remove
2. Default z-order for created components changed from 1 to 0. Otherwise redbox can't be show since it creates with default z-order wich is 0
3. Test ImageApp added for future fixes on  component. (not related to inbox, but related to my next tasks and was added to branch before redbox fixes)

Issue #23, opened 2017-08-16
"Unable to locate Ubuntu Components qmldir" error during startup


vkjr:
### Description
`Unable to locate Ubuntu Components qmldir` error appears during startup of every react-native-linux app.

### Example
<img width="1003" alt="screen shot 2017-08-16 at 6 42 01 pm" src="https://user-images.githubusercontent.com/5786310/29372016-a7422582-82b2-11e7-8605-73cc602f21dc.png">

### ToDo:
react-native-linux contains dependencies from Ubuntu qml components. They should be removed.

PR #36, opened 2017-08-19
refs #23 - Dependency from Ubuntu qml components removed

vkjr:
Dependency from Ubuntu qml components removed.

`UbuntuTextfieldManager`, `UbuntuDatepickerManager`, `UbuntuPageManager` removed because not used currently. Later they should be substituted with qt-quick-based component managers.

`UbuntuNavigatorManager`, `UbuntuScrollViewManager` were renamed and untied from Ubuntu components, but not removed, because they are referenced in javascript implementation files in `Libraries`:

- `react-native-linux/Libraries/Lists/ListView/ListView.js`
- `react-native-linux/Libraries/CustomComponents/Navigator/Navigator.ubuntu.js`


ReactQt/runtime/src/reactnavigatormanager.cpp
```diff
@@ -85,89 +85,96 @@ void UbuntuNavigatorManager::clear(int containerTag)
...
 {
   Q_ASSERT(qobject_cast<QQuickItem*>(object) != nullptr);
   return new NavigatorPropertyHandler(object);
 }
 
-QString UbuntuNavigatorManager::moduleName()
+QString ReactNavigatorManager::moduleName()
 {
-  return "UbuntuNavigatorManager";
+  return "RCTNavigatorManager";
```
MaxRis: It's might be better to start naming of modules without "RCT" prefix.
"RCT" prefix is some kind of old naming standard for react-native modules.
https://github.com/status-im/react-native-linux/blob/react-native-qt/ReactQt/runtime/src/reactmoduledata.cpp#L116

ReactQt/runtime/src/reactscrollviewmanager.cpp
```diff
@@ -125,31 +125,26 @@ void UbuntuScrollViewManager::addChildItem(QQuickItem* scrollView, QQuickItem* c
...
 import QtQuick 2.4
-import Ubuntu.Components 1.2
+import QtQuick.Controls 1.4
 
 Flickable {
-  id: flikka
-  clip: true
-  contentWidth: contentItem.childrenRect.width
-  contentHeight: contentItem.childrenRect.height
-//   Scrollbar {
```
MaxRis: For desktop apps, it might be better to display scroll bars, instead of just Flickable. We might consider this in future.

vkjr: @MaxRis, I agree. ScrollBar supported since Qt 5.7, so we can't add it in current implementation. Left for future.

ReactQt/runtime/src/reactnavigatormanager.cpp
vkjr: Agree

PR #42, opened 2017-08-23
fixes #33

### LINKED ISSUE #33: TicTacToe header text in a wrong position
### Description
TicTacToe app window has a header text `EXTREME T3` that should be at the top accordingly to javascript code:

```
return (
        <Text style={styles.title}>EXTREME T3</Text>
        <View style={styles.board}>
          {rows}
        </View>
        <GameEndOverlay
          board={this.state.board}
          onRestart={this.restartGame}
        />
      </View>
    );
```

But in react-native-qt it draws at the bottom:
screen shot 2017-08-17 at 6 25 32 pm

narek-h:
Fixed incorrect item placement. Using incremental indices to set child items.

Details: 
The actual issue of wrong placement was caused by the logic of adding the item to the layout list 
https://github.com/status-im/react-native-linux/blob/react-native-qt/ReactQt/runtime/src/reactflexlayout.cpp#L812

Adding with 0 indices results to a wrong order of items. 

on iOS, adding the children was done using similar incremental indices, though haven't found explicitly setting Z orders there, but according to this comment 
https://github.com/status-im/react-native-linux/blob/react-native-qt/ReactQt/runtime/src/reactuimanager.cpp#L180
it seems setting Z orders of sub-items is expected. 

On the other hand, having incremental Z order for children should not break any logic, since for same z order the latest child is drawn above the earlier items. 
So if the children don't have any overlap we are good, if they have, next items are drawn above previous ones.

### DIFF
--- a/ReactQt/runtime/src/reactuimanager.cpp
+++ b/ReactQt/runtime/src/reactuimanager.cpp
@@ -128,8 +128,8 @@ void ReactUIManager::setChildren
 {
   //TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
   QList<int> indices;
-  foreach(int i, childrenTags) {
-    indices.append(0); //0 is a default z-order in qml
+  for(int i = 0; i < childrenTags.size(); ++i) {
+    indices.append(i); //Incrementing per each child. Similar to iOS. See the comment in manageChildren
   }
   manageChildren(containerReactTag, QList<int>(), QList<int>(), childrenTags, indices,  QList<int>());
 }
--- a/ReactQt/runtime/src/reactuimanager.cpp
+++ b/ReactQt/runtime/src/reactuimanager.cpp
@@ -130,7 +130,7 @@ void ReactUIManager::setChildren
   //TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
   QList<int> indices;
   for(int i = 0; i < childrenTags.size(); ++i) {
-    indices.append(i); //Incrementing per each child. Similar to iOS. See the comment in manageChildren
+    indices.append(i);
   }
   manageChildren(containerReactTag, QList<int>(), QList<int>(), childrenTags, indices,  QList<int>());
 }
@@ -191,7 +191,6 @@ void ReactUIManager::manageChildren
         vm->addChildItem(container, child, i);
       } else {
         child->setParentItem(container);
-        child->setProperty(ReactItem::layout_index_property, i);
       }
 
       // Add to layout
@@ -219,7 +218,7 @@ void ReactUIManager::replaceExistingNonRootView(int reactTag, int newReactTag)
   QQuickItem* parent = ReactFlexLayout::get(item)->parentItem();
   Q_ASSERT(parent != nullptr);
 
-  int itemIndex = item->property(ReactItem::layout_index_property).toUInt();
+  int itemIndex = ReactFlexLayout::get(parent)->getChildIndex(item);
 
   manageChildren(ReactAttachedProperties::get(parent)->tag(),
                   QList<int>(),
--- a/ReactQt/runtime/src/reactuimanager.cpp
+++ b/ReactQt/runtime/src/reactuimanager.cpp
@@ -190,7 +190,7 @@ void ReactUIManager::manageChildren
         vm->addChildItem(container, child, i);
       } else {
         child->setParentItem(container);
-        child->setZ(i);
+        child->setProperty("index", i);
       }
 
       // Add to layout
@@ -218,7 +218,7 @@ void ReactUIManager::replaceExistingNonRootView(int reactTag, int newReactTag)
   QQuickItem* parent = ReactFlexLayout::get(item)->parentItem();
   Q_ASSERT(parent != nullptr);
 
-  int itemIndex = item->z();
+  int itemIndex = item->property("index").toUInt();
 
   manageChildren(ReactAttachedProperties::get(parent)->tag(),
                   QList<int>(),
--- a/ReactQt/runtime/src/reactuimanager.cpp
+++ b/ReactQt/runtime/src/reactuimanager.cpp
@@ -31,6 +31,7 @@
 #include "reactattachedproperties.h"
 #include "reactviewmanager.h"
 #include "reactuimanager.h"
+#include "reactitem.h"
 
 
 int ReactUIManager::m_nextRootTag = 1;
@@ -190,7 +191,7 @@ void ReactUIManager::manageChildren
         vm->addChildItem(container, child, i);
       } else {
         child->setParentItem(container);
-        child->setProperty("index", i);
+        child->setProperty(ReactItem::layout_index_property, i);
       }
 
       // Add to layout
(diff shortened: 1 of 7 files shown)


ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -128,8 +128,8 @@ void ReactUIManager::setChildren
 {
   //TODO: This is a simple implementation which fixes a broken example. It's not properly tested and may need revisiting
   QList<int> indices;
-  foreach(int i, childrenTags) {
-    indices.append(0); //0 is a default z-order in qml
+  for(int i = 0; i < childrenTags.size(); ++i) {
+    indices.append(i); //Incrementing per each child. Similar to iOS. See the comment in manageChildren
```
MaxRis: @narek-h , it seems we shouldn't set z-order on our side at all. (maybe for some not trivial hacks)

If JS code doesn't set some specific [zIndex](https://github.com/status-im/react-native-linux/blob/react-native-qt/Libraries/StyleSheet/LayoutPropTypes.js#L498) , we should use just 0 for zOrder, or to not set it at all, because 0 is default.

Objective-c implementation for iOS should give good example as it's done there.

narek-h: hi @MaxRis, 
thanks for the comment. I agree that setting z order ourselves (without having that set intentionally in js ) seems not natural and possibly can bring to some unexpected cases (though in my understanding it has pretty low chance). 
I tried to follow/continue the logic of existing  `manageChildren` method, where the indices of children for setting into parent was used directly to set Z order. And if I get correctly, the comment there seems to indicate that - "//on iOS, order of the subviews implies z-order, implicitly its the same in QML". 
I can remove setting Z order or set it to 0 as suggested.

MaxRis: Hi @narek-h , mentioned comment seems just means that Zposition of UIView on iOS is similar functionality as Z order in QML. But I was not able to find in iOS implementation explicit setting of Zposition in react-native implementation. This also should mean that we shouldn't do it as well.
It might be great to look into possibility to support react-native zIndex property in scope of this bug, or create separate issue for this. Adding support of zIndex and covering with some tests should resolve current issue.

MaxRis: As I understand, setting of Z order to 0 on all items for some reasons causes current issue.

vkjr: @MaxRis, @narek-h, in addition to abovementioned, setting non-zero Z-indexes will break existing redbox functionality.

narek-h: @MaxRis, right, will be good to have support of Z index of  react components, but that will not solve this issue, since we actually don't have Z index set on js part.  
current issue is not caused by Z index directly - it's caused by using same 0 values as indices to set children into layout (in my understanding and as testing showed, caused by wrong order in the list of items - PR initial description). 
I'll adjust the solution to not use Z index. 

@vkjr, this code in PR works ok for redbox case. But anyway needs to be changed, and probably the final code will have 0 Z indices.

MaxRis: @narek-h thank you
I've tried to not set Z-order for child elements at all and TicTacToe text was rendered at the top.
This makes think that exactly setting of Z-order cause the issue, but not order of items on inserting.

narek-h: @MaxRis , of course I'm gonna take a closer look on this, but I'm very curious about your solution (maybe I'm missing something). 
When I comment out setting Z order for the children and passing all 0 indices, I have same situation with text - it's drown at the bottom.

MaxRis: @narek-h , I will check this again and let you know, thanks

MaxRis: Hi @narek-h , I'm sorry, it seems I can't repro behavior that text is drawn at the top of cells with commented out setting of Z order and 0 indexes;
But if you take changes in your current PR ( indexes are incremented ) and just comment out setting of Z order, the text will drown at the top of cells.
So it seems that inserting into layout `rfl->insertChild(i, child);` in `ReactUIManager::manageChildren` reorders items from top to bottom. For me it's not clear why Z order is set here.

narek-h: @MaxRis , thanks for looking on that.  I agree it seems weird to use the index in layout to set Z orders.
Another funny thing is, if I use incremental indexes for layout insertion, and comment out setting z index in addChildItem method  ` vm->addChildItem(container, child, i);`  after pressing NewGame (second round)  the title gets lost/hidden  : ) 
I'll look on this and will try to find a solution where we don't set Z index at all, but everything is properly handled.  At this moment don't have such solution.

vkjr:
@narek-h, @MaxRis, I took a look at ios code and in file `UIView+React.m` found methods `insertReactSubview()` and `removeReactSubview()`. And I have a guess on how indices should be treated in our, Qt `UIManager`. 
Could you please help me check this guess? 


Ad I understood, `insertReactSubview`() and `removeReactSubview()` eventually called when you call `UIManager::manageChildren`.

In these methods we can see that `UIView` manages list of subviews. And when we reference index in `UIManager::manageChildren`
it eventually results in an index in that subview list.

In Qt we have similar but slightly different behavior. Every `QQuickItem` also has a list of visual children that can be obtained by `QQuickItem::childItems()`. But `QQuickItem` doesn't have methods to change order of that items. 

So it seems to me that when we '`manageChildren`' in some View (`QQuickItem`), indices mean order of child views in `QQuickItem::childItems()` list. I would treat this as 'xy-index', because in qml that order is an order of adding items to layouts.

Lets say we have ParentQQuickItem with childItems() list: `[ Child1, Child2, Child3 ].`
Indices will be: 0 for Child1, 1 for Child2, 2 for Child3.

And if we want to draw Child3 first, we should call `UIManager::manageChildren(Child3, [2], [0]....)` (move from index 2 to index 0).
As a result indices will be: 0 for Child3, 1 for Child1, 2 for Child2.

If this guess is correct, then "`indice`" of a particular child view is not a property of that child view. It is knowledge of its parent. It looks like keeping a separate `layoutIndex` property would result in a lot of work. For abovementioned example we will have to do 3 index reassignments:
```
Child3.layoutIndex = 0;
Child1.layoutIndex = 1;
Child2.layoutIndex = 2; 

```
Since `QQuickItem` doesn't have functions to change children order, probably `UIManager::manageChildren` should act like this:
1. prepare temporary list of children in correct order
2. detach all children from current item to clear `childItems()` list
3. attach all children from temporary list to create correct order in childItems()

With this approach we won't need additional property.


What do you think?

MaxRis:
@vkjr great catch!
Since `ReactFlexLayout` seems hadles the list of children items with their actual indices, it might be possible to use  `ReactFlexLayout` to remove and insert only needed items, as it's done in iOS implementation (file `RCTUIManager.m` method `_manageChildren`).

narek-h:
@vkrj, @MaxRis, 
agree with you points. A couple of clarification on current changes
. Added `layoutIinex` property does not touch the current logic of index handling, it just replaced the logic of setting Z index and using that as item index.  
In manageChildren method, the index of item was setting as z order, and in method `ReactUIManager::replaceExistingNonRootView` the index was retrieved from the item using `item->z()`.
We did not want to set Z index of the items, and to keep the current logic as is I just replaced z property with a custom one. 
.Based on your comments I found that the item index in parent layout can be retrieved without keeping the index in item (doing iteration on ReactFlexLayout->children and find the matching item). 
So I agree that this property is redundant. 
. The move functionality is not implemented yet, so anyway we had to deal with the mentioned logic. I mean, I don't think (or maybe missing something) this property complicates the things. It was introduced to replace Z property, but I agree this is redundant. 

Thanks for pointing out this. I'll remove the `layoutIndex` property.

PR #60, opened 2017-09-03
Clang format file along with 2 files modified as an example

vkjr:
### Notes
@MaxRis, @narek-h, I added `.clang_format` file with some styling. Unfotunately `clang` can't change only one setting (like tab=4 spaces). It will anyway apply a lot of changes anyway, just use default values for them. (https://stackoverflow.com/questions/30763869/how-can-i-apply-only-one-clang-format-action). 
So I generated default values file and tweaked it a bit.
You can see the formatting example in two commited files. Please tell if you are ok with these settings or would like to changes something.
Thanks.

### Installation
install clang-format: `sudo apt-get install clang-format`
enable beautifier plugin: http://doc.qt.io/qtcreator/creator-beautifier.html


MaxRis:
#61

ReactQt/runtime/src/reactimagemanager.cpp
```diff
@@ -11,348 +11,276 @@
...
-        m_bridge->enqueueJSCall("RCTEventEmitter", "receiveEvent",
-                                QVariantList{reactTag, normalizeInputEventName(eventNames[event]), data});
-      }
+    QUrl source = imageSource["uri"].toUrl();
+    if (source.isRelative())
+        source = QUrl::fromLocalFile(QFileInfo(imageSource["uri"].toString()).absoluteFilePath());
+
+    m_bridge->imageLoader()->loadImage(source, [=](ReactImageLoader::Event event, const QVariantMap& data) {
+        if (event == ReactImageLoader::Event_LoadSuccess)
+        {
```
MaxRis: @vkjr @narek-h I'm sorry for another one late feedback on formatting,
But how would you prefer if parentheses for `if conditions/loops/switches and so on` will start on the same line of code?
So it will look like:
if (event == ReactImageLoader::Event_LoadSuccess) {
    // Some code here
} else {
   // Some code here
}

Main argument is to have more valuable code in less screen space and it's the standard which is suggested , for example, by Google. Also, for example, Python language doesn't need parentheses at all (indents are used solely)

Also, can we have mandatory SPACE symbol between operators and expressions, for example:
`if`SPACE`(some_expression)`SPACE`{` ...

And also comments format, can we have it like this (SPACE symbol after // and starting with capital letter):

// Some comment here
// continue of comment if too long

Thanks

vkjr: @MaxRis, I'm ok with open brace on the same line.
Spaces in control expressions controlled by `SpaceBeforeParens: ControlStatements`, we already have it set.
Didn't find settings for spaces in comments.

PR #71, opened 2017-09-12
fixes #35

### LINKED ISSUE #35: Redbox "reload" button crashes app
In TicTacToe game if you get a Redbox screen, pressing "reload" crashes the app

narek-h:
Fixed crash on Redbox reload

Description:  
The crash was caused by the mismatch of module id and sequential index  - https://github.com/status-im/react-native-linux/blob/react-native-qt/ReactQt/runtime/src/reactbridge.cpp#L362 
On reload the list of modules was being cleared but the IDs were incremented from the last ID. 
So, on reload resetting the nextModuleId. 

Test: 
.Add a syntax error in Examples/TicTacToe/TicTacToe.js , run the example. 
=>Make sure Redbox is displayed 
. Fix the error in js file and save. 
. Press Reload on Redbox
=>Make sure TicTacToe is properly displayed.

### DIFF
--- a/ReactQt/runtime/src/reactbridge.cpp
+++ b/ReactQt/runtime/src/reactbridge.cpp
@@ -345,7 +345,7 @@ void ReactBridge::initModules() {
         ReactModuleInterface* module = qobject_cast<ReactModuleInterface*>(o);
         if (module != nullptr) {
             module->setBridge(this);
-            ReactModuleData* moduleData = new ReactModuleData(o);
+            ReactModuleData* moduleData = new ReactModuleData(o, d->modules.size());
             d->modules.insert(moduleData->id(), moduleData);
         } else {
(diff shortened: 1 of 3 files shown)


vkjr:
It is great, that we can now fix errors in runtime!

I think that when ReactBridge calls `ReactModuleData::resetModuleIdCounter`, it looks a bit like "magic knowledge", because `ReactModuleData` doesn't have `moduleIdCounter` in its interface and bridge know nothing about it, only we, developers do :)
I would suggest to add `moduleId` argument to `ReactModuleData` constructor. And when we create this object from bridge we can use `new ReactMouduleData(o, d->modules->count())` and can remove `moduleIdCounter` at all.

What do you think?

narek-h:
@vkjr, thanks. I think this is a good idea. 
I'll adjust the solution accordingly.

PR #73, opened 2017-09-13
Examples refactoring

vkjr:
Now example directories don't have *.sh and *.qml files. They are autogenerated by cmake.
Example for Button added


narek-h:
@vkjr , is Button example supposed to run properly or it's just the example part and the component is not completely supported yet ? 
I'm getting error when running the example 
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in.
```

vkjr:
@narek-h, Button not implemented yet. So far it is just an example that should work when Button implemented.

PR #79, opened 2017-09-17
Now you don't have to specify <Button> size explicitly

vkjr:
- Implicit height and width added to Button
- ReactButtonManager now instructs flex layout that implicit height and width of ReactButton.qml should be taken into account when calculating layout size.
- Default color added to Button
- Button becomes darker when pressed
- p_nativeID property added to ReactView, to make inspection in GammaRay easier

### DIFF
--- a/ReactQt/runtime/src/qml/ReactButton.qml
+++ b/ReactQt/runtime/src/qml/ReactButton.qml
@@ -7,7 +7,7 @@ Button {
 
     property string p_accessibilityLabel
     property string p_title
-    property string p_color
+    property string p_color: "#e0e0e0"
     property bool p_disabled: false
     property bool p_onPress: false
     property string p_testID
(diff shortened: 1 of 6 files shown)


ReactQt/runtime/src/qml/ReactButton.qml
```diff
@@ -17,10 +17,13 @@ Button {
...
+    implicitHeight: 40
+    implicitWidth: 100
+
     background: Rectangle {
-        anchors.fill: parent
-        color: buttonRoot.p_color
-    }
+            color: buttonRoot.down ? Qt.darker(buttonRoot.p_color, 1.2) : buttonRoot.p_color
+            visible: !buttonRoot.flat || buttonRoot.down || buttonRoot.checked || buttonRoot.highlighted
+        }
```
MaxRis: Looks like formatting issue

ReactQt/runtime/src/reactflexlayout.cpp
```diff
@@ -309,6 +309,7 @@ QDebug operator<<(QDebug debug, const ReactFlexLayoutPrivate* p) {
 
 ReactFlexLayout::ReactFlexLayout(QObject* parent) : QObject(parent), d_ptr(new ReactFlexLayoutPrivate(this)) {
     Q_D(ReactFlexLayout);
+    // qDebug() << "Layout created for: " << parent;
```
MaxRis: probably, should be removed

vkjr:
@narek-h, could you pleaser review this PR? Thanks.

Issue #75, opened 2017-09-14
Extend Integration Test with test case for expected js exception


MaxRis:
Unit tests should be able to receive notification if js exception happened on js side.
Test case to test handling of js exceptions should be added.

PR #83, opened 2017-09-18
refs #75 - Unit test JS exception; Reload to specified JS bundle

MaxRis:
- Possibility to reload ReactView with specified JS bundle URL ( One unit test class can load different JS bundles in different test methods. It's not required to create separate QML files for each different JS bundle used inside 1 test class  )
- Unit test to validate receiving of exception thrown by JS code
- Some slots titles adjusted to format `onSomethingChanged`. Prefix `on` is used to distinguish slots from signals.
- Macroses to define base implementation for unit tests `init` and `clean` functionality

### DIFF
--- a/ReactQt/runtime/src/reactredboxitem.cpp
+++ b/ReactQt/runtime/src/reactredboxitem.cpp
@@ -28,6 +28,8 @@ enum Roles {
 static QHash<int, QByteArray> roleData{
     {LineNumber, "lineNumber"}, {Column, "column"}, {File, "file"}, {MethodName, "methodName"},
 };
+
+const char REDBOX_MESSAGE_PROPERTY[] = "message";
 }
 
 class ReactRedboxItemPrivate : public QAbstractListModel {
@@ -116,7 +118,7 @@ ReactRedboxItem::~ReactRedboxItem() {}
 
 void ReactRedboxItem::showErrorMessage(const QString& message, const QList<QVariantMap>& stack) {
     Q_D(ReactRedboxItem);
-    d->redbox->setProperty("message", message);
+    d->redbox->setProperty(REDBOX_MESSAGE_PROPERTY, message);
     d->setStack(stack);
 
     QQuickItem* rootView = d->bridge->visualParent();
@@ -130,8 +132,12 @@ void ReactRedboxItem::showErrorMessage(const QString& message, const QList<QVari
 
 void ReactRedboxItem::updateErrorMessage(const QString& message, const QList<QVariantMap>& stack) {
     Q_D(ReactRedboxItem);
-    d->redbox->setProperty("message", message);
+    d->redbox->setProperty(REDBOX_MESSAGE_PROPERTY, message);
     d->setStack(stack);
+}
+
+QString ReactRedboxItem::errorMessage() const {
+    return d_func()->redbox->property(REDBOX_MESSAGE_PROPERTY).toString();
 }
 
 #include "reactredboxitem.moc"
--- a/ReactQt/runtime/src/reactview.cpp
+++ b/ReactQt/runtime/src/reactview.cpp
@@ -18,6 +18,7 @@
 #include "reactbridge.h"
 #include "reactevents.h"
 #include "reactflexlayout.h"
+#include "reactredboxitem.h"
 #include "reactuimanager.h"
 #include "reactview.h"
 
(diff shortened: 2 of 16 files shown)


ReactQt/runtime/src/reactredboxitem.cpp
```diff
@@ -134,4 +134,8 @@ void ReactRedboxItem::updateErrorMessage(const QString& message, const QList<QVa
     d->setStack(stack);
 }
 
+QString ReactRedboxItem::errorMessage() const {
+    return d_func()->redbox->property("message").toString();
```
vkjr: String "message" has 3 occurrences in the file, it would be nice to make it constant to simplify possible changes in the future.

ReactQt/runtime/src/reactview.cpp
```diff
@@ -109,8 +110,19 @@ private Q_SLOTS:
         if (bridge != nullptr && bridge->ready())
             monitorChangeUrl();
     }
+
+    void codeLocationChanged();
 };
 
+void ReactViewPrivate::codeLocationChanged() {
```
vkjr: I would suggest to use more meaningful name on slots that describe functionality, for example `resetBundleUrl()`.
For cases, when signal should result in non-related actions, slot can be called `onCodeLocationChanged()` to distinguish it from signal during textual search.

Also, in this particular case all statements work with `Bridge` object, so we can make it a slot of `Bridge` class - `Bridge::setBundle()`. Maybe we will need to invoke it from another places in the code.

ReactQt/tests/reacttestcase.cpp
```diff
@@ -67,6 +78,11 @@ void ReactTestCase::waitAndVerifyJsAppStarted() {
     waitAndVerifyCondition([=]() { return bridge()->jsAppStarted(); }, "js app not started, failed by timeout");
 }
 
+void ReactTestCase::waitAndVerifyJSException(const QString& exceptionMessage) {
+    waitAndVerifyCondition([=]() { return bridge()->redbox()->errorMessage() == exceptionMessage; },
+                           "Expected JS exception was not received");
```
vkjr: Adding expected exception message to output can simplify understanding the error from console.

ReactQt/tests/test-integration.cpp
```diff
@@ -25,30 +27,36 @@ private slots:
     void cleanupTestCase();
 
     void testTestModuleMarkTestCompleted();
+    void testJSExceptionReceived();
 };
 
 void TestIntegration::initTestCase() {
```
vkjr: `initTestCase()` and `cleanupTestCase()` can be omitted, since they are virtual.

vkjr:
@MaxRis, could you please check Circle output. It looks like builds fails because test uses ReactRedboxItem.qml that still depends on old `QtQuick.Controls 1.4`, that is not installed in test environment. Probably we need to move ReactRedboxItem.qml to latest `QtQuick.Controls` version.

MaxRis:
Thank you @vkjr to find out above things, I'm still looking into the failed unit tests. Will notify once the final version of changes will be ready, thanks

ReactQt/tests/test-integration.cpp
MaxRis: @vkjr , that's a good point, but it seems QTest requires these methods to be defined explicitly in each class passed to QTEST_MAIN macro.
I was not able to get base class virtual methods called.
And more over, it seems, others test cases has similar issue, that `cleanupTestCase` is not get called at all (since `initTestCase()` is redefined, but  `cleanupTestCase` not ).

MaxRis: My best guess, is that QTest lib just defines it's own version of `initTestCase()` and `cleanupTestCase()`  if they are not defined in class.

vkjr: @MaxRis, interesting info, thanks!

ReactQt/tests/reacttestcase.h
```diff
@@ -10,23 +10,36 @@
...
 
+#define INIT_TEST_CASE_DEFAULT(BaseClass)                                                                              \
+    virtual void initTestCase() override {                                                                             \
+        BaseClass::initTestCase();                                                                                     \
+    }
+
+#define CLEANUP_TEST_CASE_DEFAULT(BaseClass)                                                                           \
+    virtual void cleanupTestCase() override {                                                                          \
+        BaseClass::cleanupTestCase();                                                                                  \
+    }
```
MaxRis: @vkjr have added macros to define default implementation of init and cleanup unit tests' methods

MaxRis:
@vkjr @narek-h have updated description of PR, now, finally, it's ready for review, thank you!

MaxRis:
@narek-h could you please check the changes for this PR as well, thank you

Issue #87, opened 2017-09-20
Implement <Slider> component support


MaxRis:
[Slider](https://facebook.github.io/react-native/docs/slider.html) component support should be implemented.

vkjr:
@MaxRis, can we consider this issue as `done`?

MaxRis:
@vkjr not really, not all props are implemented. Also it contains child ticket with things to implement.

PR #90, opened 2017-09-24
refs #87 - Basic Slider component props support; Slider props unit test

MaxRis:
- Basic support of Slider component (basic props)
- Slider props unit test
- Added SliderExample.js to the list of examples supported by RnTester

### DIFF
--- a/ReactQt/tests/test-slider-props.cpp
+++ b/ReactQt/tests/test-slider-props.cpp
@@ -14,11 +14,8 @@
 #include <QTest>
 #include <QtQuick/QQuickView>
 
-class TestButtonProps : public ReactPropertyTestCase {
+class TestSliderProps : public ReactPropertyTestCase {
     Q_OBJECT
-
-private:
-    QQuickItem* qmlImage();
 
 private slots:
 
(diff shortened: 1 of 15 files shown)


ReactQt/tests/test-slider-props.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+void TestSliderProps::initTestCase() {
+    ReactPropertyTestCase::initTestCase();
+    loadQML(QUrl("qrc:/TestSliderProps.qml"));
+    waitAndVerifyJsAppStarted();
+}
+
+QVariantMap TestSliderProps::propValues() const {
+    return {{"p_value", 0.5},
```
vkjr: Probably it would be nice to cover here all the properties we have in Slider component qml.

MaxRis: @vkjr initially I've added to Slider.qml the whole list of props from Slider.js, and finally supposed to extend props unit test only with fully supported props. Others than currently listed props are not really supported, I hope to implement their support in next commits, so will extend unit test also.

vkjr: @MaxRis, ok I see. I usually consider this "props" tests just as a proof that property value passes well between JS and QML, since they don't check any functionality.

MaxRis: @vkjr , got it, thanks, Have updated to cover all QML properties

PR #95, opened 2017-09-26
React native linux switched to Yoga layout engine

vkjr:
### Changes
- Old layout engine removed, `Yoga` engine added instead
- `ReactView.h\cpp` renamed to RootView to not confuse with ReactView.qml

### How `Yoga` engine works
`Yoga` engine has api that works with "nodes". Every `YGNode` has a set of properties. Some of them used to setup layout settings, some - to read the calculated position of the node.
When 

Every `YGNode` has a set of properties that control the layout of node (like minWidth, alignment, justification, etc.) and set of properties that contain calculated layout results (like x, y, width, height). You can join nodes in hierarchy and `calculate` it. `Yoga` engine will read "controlling" properties and save the result in "position" properties.

### How `Yoga` engine used in `React Native Linux`
A lot of components in `React Native` have layout properties that are almost identical to properties in `Yoga`.
We have a `Flexbox` object that is wrapper to `YGNode`. It is responsible for getting layout props from JS side and turning them into QML position properties. Here is the process:
1. We receive a set of layout properties from JS side.
2. `ReactNativeHander` search for `Flexbox` object for control and set all the layout properties to `Flexbox`.
3. Whenever `Flexbox` gets a property update, it passes the value to internal `YGNode` object.
4. Whenever `UIManager` receives update on component hierarchy (add/delete/move), it also requests `Flexbox` to update hierarchy of `YGNode`s
5. When `RootView` is ready, it triggers recalculation of `Flexbox` hierarchy
6. When calculation completed, `Flexbox` read results from `YGNode`s and assign them to QML controls.

### Notes
@MaxRis, @narek-h, could you please take a look at `flexbox.cpp` It contains delimiters between functions, like this:
```
void funcA() {}
//---------------------------
void funcB() {}
```

For me these lines significantly improve readability of cpp file. I clearly see where function starts/ends.
What do you think about the delimieters? Would you find them usable?


ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -164,31 +155,38 @@ void ReactUIManager::manageChildren(int containerReactTag,
...
     }
 
-    QQuickItem* parent = ReactFlexLayout::get(item)->parentItem();
+    QQuickItem* parent = oldItem->parentItem();
     Q_ASSERT(parent != nullptr);
 
-    int itemIndex = ReactFlexLayout::get(parent)->getChildIndex(item);
+    int itemIndex = -1;
+    for (int i = 0; i < parent->childItems().count(); ++i) {
+        if (oldItem == parent->childItems().at(i)) {
```
MaxRis: Looping across all items might affect performance ... with TicTacToe example I got loops with 2-3 iterations, but it seems the number of iterations can be any based on input JS. Maybe, makes sense to create ticket to address this later.

ReactQt/runtime/src/rootview.cpp
```diff
@@ -112,96 +110,100 @@ private Q_SLOTS:
...
+    Q_D(RootView);
 
     setAcceptedMouseButtons(Qt::LeftButton);
     setFiltersChildMouseEvents(true);
 
     d->bridge = new ReactBridge(this);
     connect(d->bridge, SIGNAL(readyChanged()), SLOT(bridgeReady()));
 
-    connect(this, SIGNAL(liveReloadChanged()), d, SLOT(onLiveReloadChanged()));
+    connect(this, SIGNAL(liveReloadChanged()), d, SLOT(liveReloadChanged()));
```
MaxRis: I'm getting warning on runtime
`QObject::connect: No such slot RootViewPrivate::liveReloadChanged() in ..src/rootview.cpp:122`

ReactQt/runtime/src/rootview.cpp
```diff
@@ -232,8 +234,14 @@ void ReactView::bridgeReady() {
     }
 }
 
-void ReactView::componentComplete() {
-    Q_D(ReactView);
+void RootView::requestPolish() {
+    if (ReactAttachedProperties::get(this)->tag() == -1)
```
MaxRis: It might make sense to wrap this check into method with easy to ready and understand title.

ReactQt/runtime/src/reactuimanager.cpp
```diff
@@ -111,43 +95,50 @@ void ReactUIManager::setChildren(int containerReactTag, const QList<int>& childr
...
+
+        QList<QQuickItem*> itemsToRemove;
+        for (int i : removeAtIndices) {
+            itemsToRemove.push_back(parent->childItems().at(i));
+        }
+
+        for (QQuickItem* child : itemsToRemove) {
+            child->setParent(0);
+            m_views.remove(ReactAttachedProperties::get(child)->tag());
+            delete child;
```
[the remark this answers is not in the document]
vkjr: I had a bug - QQuickItem was removed from its parent when it shouldn't be. And I couldn't figure out when this happens because deleteLater() was used and stacktrace was unclear. So I substitute deleteLater() with direct deletion to prevent similar problems in future. But can get it back.

ReactQt/runtime/src/reactitem.cpp
```diff
@@ -558,8 +558,9 @@ void ReactItem::paint(QPainter* painter) {
 }
 
 void ReactItem::updatePolish() {
-    ReactFlexLayout::get(this)->setWidth(width());
-    ReactFlexLayout::get(this)->setHeight(height());
-
-    ReactFlexLayout::polish(this);
+    if (childItems().count() > 0) {
+        QQuickItem* view = childItems().at(0);
```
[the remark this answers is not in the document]
vkjr: This method should be moved to RootView, since we recalculate layout from top node. RootView usually have only one child, I'll add assert to check this.

ReactQt/runtime/src/reactpropertyhandler.cpp
```diff
@@ -36,17 +43,17 @@ void ReactPropertyHandler::applyProperties(const QVariantMap& properties) {
...
-        if (it != m_HandlerProperties.end()) {
+
+        auto it = m_qmlProperties.find(key);
+        if (it != m_qmlProperties.end()) {
+            QMetaProperty property = it.value();
+            setValueToObjectProperty(m_object, property, propertyValue);
+        }
+
+        it = m_flexboxProperties.find(key);
+        if (it != m_flexboxProperties.end() && m_flexbox) {
```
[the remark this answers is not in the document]
vkjr: Currently `ReactItem` and `Flexbox` have duplicated properties (like borderWidth). `Flexbox` properties should be assigned anyway to make layout calculation correct.
Probably we can get rid of duplicated properties in `ReactItem`, I will make this as an improvement later. Added #100 for it.

ReactQt/runtime/src/reactuimanager.cpp
vkjr: Replaced loop with `indexOf()` function. that looks better, but don't think that it is faster.
I would suggest to create a ticket on this if we run into performance issues. So far, hardly can imagine UI where we have so many children that it impacts performance.

ReactQt/runtime/src/rootview.cpp
vkjr: Thanks for noticing. Incorrectly merged.

vkjr:
@MaxRis, thanks for all findings!

Fixed notes and removed delimiters.

ReactQt/runtime/src/reactuimanager.cpp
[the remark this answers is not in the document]
MaxRis: @vkjr  I have found following so far: http://doc.qt.io/qt-5/qquickitem.html
```
Note: Use of QObject::deleteLater() to clean up graphics resources is not recommended as this will run at an arbitrary time and it is unknown if there will be an OpenGL context bound when the deletion takes place.
```
Seems better to follow your implementation :)

ReactQt/runtime/src/reactuimanager.cpp
MaxRis: @vkjr now it looks nicer at least! Thank you
We might discover such UI pretty soon :)

PR #105, opened 2017-10-02
Refs #86 Basic TextInput

narek-h:
Added basic TextInput component with basic properties and events support.
Added RNTester example showing supported functionality. 

Note: Sometimes noticing a big delay on updating a text view for the occurred event.  Continue to type/select to see the updates. 
(probably the issue is not relevant to this patch)

### DIFF
--- a/RNTester/js/TextInputExample.js
+++ b/RNTester/js/TextInputExample.js
@@ -1,0 +1,258 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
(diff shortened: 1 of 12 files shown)


RNTester/js/TextInputExample.js
```diff
@@ -0,0 +1,258 @@
+/**
```
MaxRis: It might make more sense to name file TextInputExample.ubuntu.js since TextInput implementation for desktop differs from implementation for mobile.
Also copyright might be replaced for Status.

narek-h:
@maxr, thanks for the comments. Will add a patch addressing them,  
Also, will add unit tests by a separate PR.

narek-h:
@MaxRis, @vkjr , added  `js/utils.js` file containing the function used in qml. Later we can rename/split them to more specific files if needed. 
Addressed the other comments as well.  If everything looks ok, will merge.

MaxRis:
@narek-h looks great! Thank you!

PR #108, opened 2017-10-05
Added basic support for <Modal> component

vkjr:
- Added support for `<Modal>` component
- `Flexbox::findFlexbox()` now takes flexbox from component property. Previous implementation didn't work well for Modal. Because its flexbox in fact tied to `Popup.contentItem` component, not the root component of `ReactModal`.


ReactQt/runtime/src/qml/ReactModal.qml
```diff
@@ -0,0 +1,38 @@
...
+import QtQuick.Controls 2.2
+import QtGraphicalEffects 1.0
+import React 0.1 as React
+
+Item {
+
+    id: modalRoot
+    objectName: "modalRoot"
+
+    property string p_animationType
```
narek-h: @vkjr , p_animationType, p_transparent are not implemented yet, right ? Just dummy properties so far ? 
sorry if missing something.

vkjr: @narek-h, yes, you are right.

vkjr: @narek-h, thanks! Added record to CMakeLists.txt

Issue #116, opened 2017-10-08
Implement <Picker> component 


narek-h:

narek-h:
Current implementation of Picker uses items from old `QtQuick Controls`. We need to adjust the solution to use only new controls - `Controls 2`.

PR #122, opened 2017-10-10
Refs #116 Implemented basic Picker

narek-h:
added props unit test

Note: 
In android Picker has two types - dialog and dropdown (iOS seems to have one). 
Here added one type based on QML Tumbler

### DIFF
--- a/Libraries/Components/Picker/Picker.js
+++ b/Libraries/Components/Picker/Picker.js
@@ -15,6 +15,7 @@
 var ColorPropType = require('ColorPropType');
 var PickerIOS = require('PickerIOS');
 var PickerAndroid = require('PickerAndroid');
+var PickerUbuntu = require('PickerUbuntu');
 var Platform = require('Platform');
 var React = require('React');
 const PropTypes = require('prop-types');
@@ -162,6 +163,8 @@ class Picker extends React.Component {
      } else if (Platform.OS === 'android') {
        // $FlowFixMe found when converting React.createClass to ES6
        return <PickerAndroid {...this.props}>{this.props.children}</PickerAndroid>;
+     } else if (Platform.OS === 'ubuntu') {
+       return <PickerUbuntu {...this.props}>{this.props.children}</PickerUbuntu>;
      } else {
        return <UnimplementedView />;
      }
--- a/ReactQt/runtime/src/qml/ReactPicker.qml
+++ b/ReactQt/runtime/src/qml/ReactPicker.qml
@@ -1,0 +1,46 @@
+import QtQuick 2.7
+import QtQuick.Controls 2.2
+import React 0.1 as React
+import QtQuick.Extras 1.4
+import QtQuick.Controls.Styles 1.2
+
+
+Tumbler {
+    id: tumbler
+    property var pickerManager: null
+    property variant flexbox: React.Flexbox {control: tumbler}
+    property variant p_items: []
+    property int p_selected:0
+    property bool p_onValueChange: false
+    property string p_testID
+
+    objectName: p_testID
+
+    TumblerColumn {
+        model: p_items
+        onCurrentIndexChanged: {
+            if (pickerManager) {
+                pickerManager.sendValueChangeToJs(tumbler, currentIndex);
+            }
+        }
+    }
+
+    style: TumblerStyle {
+        id: tumblerStyle
+
+        delegate: Item {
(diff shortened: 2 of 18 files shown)


Libraries/Components/Picker/Picker.js
```diff
@@ -15,6 +15,7 @@
 var ColorPropType = require('ColorPropType');
 var PickerIOS = require('PickerIOS');
 var PickerAndroid = require('PickerAndroid');
+var PickerUbuntu = require('PickerUbuntu');
```
MaxRis: If implemented Picker control is not intended to be used only on Ubuntu, it makes sense to remove `Ubuntu` from control title. Maybe, `PickerQt` should work better?

ReactQt/runtime/src/qml/ReactPicker.qml
```diff
@@ -0,0 +1,46 @@
...
+import QtQuick.Extras 1.4
+import QtQuick.Controls.Styles 1.2
+
+
+Tumbler {
+    id: tumbler
+    property var pickerManager: null
+    property variant flexbox: React.Flexbox {control: tumbler}
+    property variant p_items: []
+    property int p_selected:0
```
vkjr: <Picker> component in React Native doesn't have `selected` property, it is called `selectedValue`. I would suggest to remove `p_` prefix because it is used to match original React Native properties by name.
Also it would be nice to add `p_selectedValue` property because without it we can't tie `Picker` value to another variable in code.

ReactQt/runtime/src/qml/ReactPicker.qml
```diff
@@ -0,0 +1,46 @@
...
+                text: styleData.value.label
+                color: styleData.value.color ? styleData.value.color : "black"
+                opacity: 0.4 + Math.max(0, 1 - Math.abs(styleData.displacement)) * 0.6
+                anchors.centerIn: parent
+            }
+        }
+    }
+
+    onP_selectedChanged: {
+        tumbler.setCurrentIndexAt(0, p_selected);
```
vkjr: Won't be called, since no one changes `p_selected` property.

Libraries/Components/Picker/Picker.js
MaxRis: @vkjr I would prefer to make separate commit to change `ubuntu` platform to something else, but all others changes (like renaming of exported classes, etc) better to make in separate commits also. It should be more easy to fix broken things which should occur with high possibility :)
Changing of `ubuntu` platform evolves updating of JS code to init and build react-native apps. This brings additional complexity.

vkjr: @MaxRis, sure, agree about different git commits :) I used incorrect words.
What I mean is that if we decide to rename `SomethingUbuntu` to `SomethingQt` in JS code, then it would be nice to do similar changes as a one activity (and probably different commits).
Quick search shows that there is also `UbuntuPage` and `UbuntuScrollView`. I think it would be nice to rename them too.
Of course it is not a big deal to have `SomethingUbuntu` and `SomethingQt` in code simultaneously, it just "smells" worse, as for me :) 

Added #125 to not forget this chages.

narek-h:
@vkjr , @Maxr, thanks for the comments. 
.as I understand (from the android component codes Libraries/Components/Picker)  `selectedValue` property is used to set the initial (default) selection. 
Then it's converted to  'selected' property internally, that's why QML side takes `selected`. 
And current logic actually works for initial selection. 
Maybe it makes sense to rename it? I took Android component and did minimal adjustments to work for our purpose.  
The part (Libraries/Components/Picker/PickerUbuntu.ubuntu.js) can be cleaned up , especially if we don't need 'dropdown picker' yet. 

. Current `PickerExample.ubuntu` is  taken from the original `PickerExample`, I just removed most part since the cases did not make much sense with current implementation  (there was disabled, colored, dropdown cases. I tried to combine in one item all the supported things). 

. Can you please confirm ubuntu renaming should be done in the context of other task? Or need to fix relevant files here ?

MaxRis:
@narek-h , since we have ticket to revisit that, it might have sense to collect files or classes to rename in comments to that ticket, thanks

vkjr:
@narek-h, you are right about `selectedValue`, sorry, I missed the internal conversion of this property.

narek-h:
Ok, thanks. Will add the items to rename to the appropriate task and merge this.

narek-h:
@maxr, @vkjr , added a small circleci fix(second commit)  to include modules needed for Picker.

Issue #117, opened 2017-10-08
Repair generation of custom react-native apps for Ubuntu platform


MaxRis:
After merge with 0.48 of origin react-native, creation of custom react-native apps for Ubuntu platforms is broken.
Command `react-native ubuntu` doesn't generate Linux port related files and directories.
Command `react-native run-ubuntu` doesn't recognized and doesn't build app sources/launch built binary.

PR #124, opened 2017-10-11
refs #117 - Repair generation and compilation of custom react-native app

MaxRis:
- Custom react-native application (file path ReactQt/application/src) is added as subdirectory to main cmake build file. Compilation of it should help to avoid future compilation issues of custom apps on significant source code updates.
- Custom app uses `Q_INIT_RESOURCE(react_resources);` to force linking of resources from react-native library http://doc.qt.io/qt-5/qdir.html#Q_INIT_RESOURCE
- Repaired creation of custom react-native app with running of commands (react-native-cli from our repo should be installed globally): 1. `react-native init ProjectName` 2. `cd ProjectName` 3. `react-native ubuntu` 4. `react-native run-ubuntu` . Old generation mechanism is restored with minor changes and tweaks ( commands description updated, automatic start of `node ubuntu-server.js`, automatic applying of patch to extend metro-bundler with `ubuntu` platform ).
- Resources references changed to URLs with `qrc:` at the start for consistency across the project. Please let me know if you are fine with it.

### DIFF
--- a/react-native-cli/index.js
+++ b/react-native-cli/index.js
@@ -248,6 +248,7 @@ function createProject(name, options) {
       start: 'node node_modules/react-native/local-cli/cli.js start',
       ios: 'react-native run-ios',
       android: 'react-native run-android',
+      ubuntu: 'react-native run-ubuntu'
     }
   };
   fs.writeFileSync(path.join(root, 'package.json'), JSON.stringify(packageJson));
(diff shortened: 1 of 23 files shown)


react-native-cli/index.js
```diff
@@ -257,7 +258,7 @@ function createProject(name, options) {
 }
 
 function getInstallPackage(rnPackage) {
-  var packageToInstall = 'https://github.com/status-im/react-native-linux#react-native-qt'
+  var packageToInstall = 'https://github.com/MaxRis/react-native-qt#bug/repair-run-ubuntu-#117';
```
MaxRis: This line should be reverted back before merge. Left this as is if someone will want to test changes locally.

PR #128, opened 2017-10-16
Touchable onLongPress fixed. Added test to cover different cases of t…

vkjr:
### Bugfix description

There was an issue with `onLongPress` event processing in `Touchable`. It never arrived while user holds mouse button pressed. It appeared that `NetExecutor` wasn't able to read more than one package of data per read and sometimes that led to data skipping.
Testcase added to cover current bug and few more scenarios of passing data via socket.

### Note
`PipeExecutor` not checked/fixed yet.


ReactQt/tests/test-netexecutor-socket.cpp
```diff
@@ -0,0 +1,188 @@
...
+    QCoreApplication::processEvents();
+}
+
+void TestNetExecutorSocket::sendPackageInChunks(quint32 bytesCount) {
+    QByteArray ba = package(bytesCount);
+
+    while (ba.length()) {
+        QByteArray chunk = ba.left(CHUNK_SIZE);
+        ba.remove(0, chunk.length());
+        m_socket->write(chunk);
```
MaxRis: `m_socket` probably should be flushed each time when you expect that data should be really sent.
To simulate situation to put multiple packets into socket buffer, maybe, make sense to extend methods "sendPackageXXX" with additional input argument `bool flushSocket`

vkjr: @MaxRis, explicit call to `QApplication::processEvents()` has the same effect as socket flushing, since data sent to network whenever app process event loop. Probably it would be more clear to call `flush()` explicitly instead, thanks.
I think that additional argument won't be helpful since I want socket to be flushed every time I sent test data.

PR #130, opened 2017-10-20
Bug/pipe communication

vkjr:
Fixed issue with communication via pipe.
Communication is done via `ubuntu-server.js` output channel. But if executed Js code has writes messages via `console.log`, that also goes into output channel and breaks messaging. To fix this we now have a line in `ubuntu-server.js`, that redirects such output to error channel `console.log = console.error`.

Executors were refactored to eliminate copypasted code.


narek-h:
@vkjr  could you please let me know how this can be tested. I supposed with pipe executor we don't need to run ubuntu-server.js manually, but I think I'm wrong.  

Do I have to modify the code to run in 'pipe' mode ?

MaxRis:
@narek-h you can try manually edit following line https://github.com/status-im/react-native-linux/blob/react-native-qt/ReactQt/runtime/src/bridge.cpp#L114 to the:

```const int executorType = QMetaType::type(("PipeExecutor").toLocal8Bit());```

vkjr:
@narek-h, To check pipe connection you need to change from `RemoteServerConnection` to `LocalServerConnection` in this line:
https://github.com/status-im/react-native-linux/blob/34631af764b0791c04909a675d288e831c37cfb7/ReactQt/runtime/src/rootview.cpp#L82

But it looks like I introduced bug in last moment and this way doesn't work. Thanks for finding it!

vkjr:
@narek-h, fix commited, please check.
Thanks!

narek-h:
Thanks guys!  Works fine.

PR #142, opened 2017-11-15
refs #7 - Developer menu and Chrome Dev Tools support

MaxRis:
- Developer menu functionality is compiled under RCT_DEV define and by default available only in Debug builds.
- Developer menu is loaded dynamically from DevMenu.qml as direct child of top root QML item.
- CTRL+R shortcut is used to show developer menu
- WebSocketExecutor class is added to communicate between native app and Chrome Dev Tools via packager (similar to RCTWebSocketExecutor.m on iOS ); IExecutor interface is back to provide similar interface for executors.
- Chrome browser seems should be installed.
- Debugging workflow: 1. Start app built in Debug. 2. Press CTRL+R and Developer menu should appear. 3. Select "Remote JS Debugging". 4. Chrome Browser tab will be opened automatically and finally status of debugging session in Browser tab should change to active (something like Status: Debugger session #10001 active.). 5. Now it's possible to press CTRL+SHIFT+J in Chrome Browser Tab, select Sources in appeared developers panel, browse to JS sources of application to put breakpoints, etc.

### DIFF
--- a/ReactQt/runtime/src/communication/websocketexecutor.cpp
+++ b/ReactQt/runtime/src/communication/websocketexecutor.cpp
@@ -1,0 +1,201 @@
+
+/**
+ * Copyright (c) 2017-present, Status Research and Development GmbH.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#ifdef RCT_DEV
+
+#include "websocketexecutor.h"
+
+#include <QDebug>
+#include <QJsonArray>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QNetworkRequest>
+#include <QQueue>
+#include <QTimer>
+#include <QWebSocket>
+
+namespace {
+const int CHROME_DEV_TOOL_AWAIT_TIMEOUT = 5000;
+}
+
+class WebSocketExecutorPrivate : public QObject {
+public:
+    WebSocketExecutorPrivate(WebSocketExecutor* e) : q_ptr(e) {}
+
+    bool processRequests();
+    void processRequest(const QByteArray& request,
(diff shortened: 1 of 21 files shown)


ReactQt/runtime/src/communication/websocketexecutor.cpp
```diff
@@ -0,0 +1,197 @@
...
+                    QByteArray data = getMessageData(QVariantMap{{"method", "prepareJSRuntime"}});
+                    processRequest(data, [=](const QJsonDocument&) {
+                        m_readyToSendAppScriptData = true;
+                        q_ptr->executeApplicationScriptOnSocketReady();
+                    });
+                    timer->deleteLater();
+                });
+                timer->start();
+            }
+            reply->deleteLater();
```
vkjr: From Qt documetation about QNetworkReply::readyRead: `This signal is emitted once every time new data is available for reading from the device's current read channel`.
That could result in double deletion of `reply` and `networkAccessManager`

ReactQt/runtime/src/rootview.cpp
```diff
@@ -288,6 +309,25 @@ void RootView::sendMouseEvent(QMouseEvent* event, const QString& eventType, QQui
     event->setAccepted(true);
 }
 
+#ifdef RCT_DEV
+void RootView::loadDevMenu() {
+    QQmlComponent component(qmlEngine(this));
```
vkjr: This chunk of code is similar to `ViewManager::createView()`. Probably we can move common function to `utilities.h`

MaxRis:
@vkjr , thank you! Have resolved found things

vkjr:
@MaxRis, it is very cool to see call stack and all debugging in chrome, thanks for this feature! :)

ReactQt/runtime/src/uimanager.cpp
```diff
@@ -384,10 +384,21 @@ UIManager::~UIManager() {
 }
 
 void UIManager::reset() {
+    // Avoid to delete root view on reset
+    QQuickItem* rootView = m_views[m_rootTag];
```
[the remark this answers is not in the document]
MaxRis: My bad, thank you for finding this out!

Issue #131, opened 2017-10-23
Support of 3rd party react-native modules installed into node_modules


MaxRis:
Build script should be able to recognize 3rd party react-native modules with linux port support installed in node_modules directory.
Straightforward way to implement this, seems, to extend cmake build script with dynamic lookup and including of found directories (inside node_modules folder) containing CMakeList.txt.

PR #147, opened 2017-11-25
refs #131 - External modules support

MaxRis:
- Application based on react-native-desktop can be extended with external native modules by adding array of external modules paths to package.json (Custom variable `desktopExternalModules`)
- `react-native run-desktop` command process `desktopExternalModules` from package.json and passes values to CMake build script.
- CMake build script updated to support EXTERNAL_MODULES_DIR input param with list of external modules paths. Each external module directory should contain its own CMakeLists.txt script which extends REACT_NATIVE_DESKTOP_EXTERNAL_MODULES_SRC variable with corresponding module C++ source files and also extends REACT_NATIVE_DESKTOP_EXTERNAL_MODULES_TYPE_NAMES variable with external modules C++ classes type names.
- List of external modules can be passed to RootView instance directly in QML code using property `externalModules`
- Each C++ native module class should be inherited from QObject and from ModuleInterface. It should register itself within Qt metadata system and has default constructor defined with Q_INVOKABLE.
- Bridge instantiate C++ classes based on their type names received from `externalModules` property. Since instantiating based on QMetaType system, C++ native module classes should meet requirements described above.

----------------------------------------------

There is available test branch with added ExternalTestModule and ExternalTestModule2: https://github.com/MaxRis/react-native-desktop/tree/feature/test-3rd-party-modules-support-%23131

package.json of custom application should be extended with
```
,
  "desktopExternalModules": [
    "node_modules/react-native/ExternalTestModule",
    "node_modules/react-native/ExternalTestModule2"
  ]
```

JS code can reference and call external modules as following:

```
var ReactNative = require('react-native');
var { ExternalTestModule, ExternalTestModule2 } = ReactNative.NativeModules;
...
  render() {
    return (
        <Text style={styles.welcome}>
          Welcome to React Native!
          {ExternalTestModule.sayHello}
          {ExternalTestModule2.sayHello}
        </Text>
      </View>
    );
  }
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,13 @@ set(PROJECT_VERSION "0.0.1")
 include_directories(./React/Layout)
 enable_testing()
 
+message(STATUS "External modules dirs: ${EXTERNAL_MODULES_DIR}")
+
+foreach(module_directory ${EXTERNAL_MODULES_DIR})
+  message(STATUS "Add external module subdirectory: ${module_directory}")
+  add_subdirectory(${module_directory})
+endforeach(loop_var)
+
 add_subdirectory(ReactQt)
 add_subdirectory(Examples)
 add_subdirectory(RNTester)
--- a/Examples/Example.qml.in
+++ b/Examples/Example.qml.in
@@ -11,5 +11,6 @@ Rectangle {
 
     moduleName: "${EXAMPLE_NAME}App"
     codeLocation: "http://localhost:8081/Examples/${EXAMPLE_NAME}/${EXAMPLE_NAME}App.bundle?platform=desktop&dev=true"
+    externalModules: []
   }
 }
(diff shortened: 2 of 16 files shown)


vkjr:
@MaxRis, thank you for implementing that, looks great!
I'm wondering, if it possible to have implementation without additional section in `package.json`? How this works for ios and android that we don't need to explicitly list them in `package.json`?

MaxRis:
@vkjr I might be wrong, but for Android is used Gradle and for iOS pods are used.
It seems modules detection is done automatically by used build systems and tools.

>  if it possible to have implementation without additional section in package.json?
Maybe some automatic scanning of all subdirs by CMake and if some specific info is found the module is automatically added. Or such scanning might be done by Node during run of `react-native run-desktop` command and based on `dependencies` from `package.json`

PR #149, opened 2017-11-28
refs #127  Basic WebView implementation

narek-h:
TEST: 
Basic functionality in 'Simple browser'  is supported - Back,Forward, Go(reload)


Libraries/Components/WebView/WebView.js
```diff
@@ -0,0 +1,416 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @providesModule WebView
```
MaxRis: @narek-h Do we really need WebView.js since packager eats only WebView.desktop.js finally?

ReactQt/runtime/src/componentmanagers/webviewmanager.cpp
```diff
@@ -0,0 +1,72 @@
...
+bool WebViewManager::shouldLayout() const {
+    return true;
+}
+
+QList<ModuleMethod*> WebViewManager::methodsToExport() {
+    return QList<ModuleMethod*>{};
+}
+
+void WebViewManager::reload() {
+    emit s_reload();
```
MaxRis: seems more safe to use Q_EMIT macro instead (not all platforms supported by Qt supports signals)

ReactQt/runtime/src/componentmanagers/webviewmanager.h
```diff
@@ -0,0 +1,50 @@
...
+
+    virtual ViewManager* viewManager() override;
+    virtual QString moduleName() override;
+    virtual bool shouldLayout() const override;
+
+    Q_INVOKABLE void reload();
+    Q_INVOKABLE void goBack();
+    Q_INVOKABLE void goForward();
+
+signals:
```
MaxRis: Q_SIGNALS

ReactQt/runtime/src/qml/ReactWebView.qml
```diff
@@ -0,0 +1,22 @@
...
+    property var p_source: ''
+    property var webViewManager: null
+    property var flexbox: React.Flexbox {control: webViewRoot}
+    url: (p_source && p_source.uri) ? p_source.uri : ""
+
+    onWebViewManagerChanged: {
+        if (!webViewManager) {
+            return;
+        }
+        webViewManager.s_reload.connect(webViewRoot.reload);
```
MaxRis: should `webViewManager` property be checked for null before used?

Libraries/Components/WebView/WebView.js
narek-h: @MaxRis , you are right. I left that somehow. Will be removed.

ReactQt/runtime/src/componentmanagers/webviewmanager.cpp
narek-h: thanks, will be fixed!

ReactQt/runtime/src/qml/ReactWebView.qml
narek-h: @MaxRis ,  agree, it seems logical that initially it's null then when it's initialized onChange is invoked. But without this check I'm getting 
`Cannot read property 's_reload' of null`
Seems somehow onChange is invoked but the property is still null.

MaxRis: @narek-h , I'm sorry, my comment is not relevant here, please, ignore

PR #158, opened 2017-12-09
lein-figwheel support

MaxRis:
- RCTNetworking updated to be compatible with react-native's API
- Dev menu extended with Hot Reloading option. ( Placeholder for HMRLoadingView is added in this scope )
- WebSocketModule native module implementation
-  Supported examples are extended with WebSocketExample
- Custom react-native desktop app is set to be built in Debug by default to enable Dev menu

### DIFF
--- a/ReactQt/runtime/src/rootview.cpp
+++ b/ReactQt/runtime/src/rootview.cpp
@@ -24,11 +24,14 @@
 
 using namespace utilities;
 
+namespace {
+
 const QString TOUCH_START = "touchStart";
 const QString TOUCH_END = "touchEnd";
 const QString TOUCH_MOVE = "touchMove";
 
-namespace {
+const int RESET_CONTENT_HTTP_STATUS_CODE = 205;
+
 QVariantMap makeReactTouchEvent(QQuickItem* item, QMouseEvent* event) {
     const QPointF& lp = event->localPos();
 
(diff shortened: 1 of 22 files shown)


ReactQt/runtime/src/rootview.cpp
```diff
@@ -93,25 +95,33 @@ class RootViewPrivate : public QObject {
...
+                    qCritical() << __PRETTY_FUNCTION__ << "Error monitoring change url";
+                    return;
+                }
+
+                qDebug() << "/onchange live reload reply received!!!";
+
+                if (!liveReload)
+                    return;
+
+                if (replyStatusCode == 205) {
```
vkjr: It would be nice to substitue constant 205 with some meaningful named constant.

ReactQt/runtime/src/websocketmodule.cpp
```diff
@@ -0,0 +1,151 @@
...
+    }
+};
+
+void WebSocketModule::connect(const QUrl& url,
+                              const QVariantList& protocols,
+                              const QVariantMap& options,
+                              qlonglong socketId) {
+    Q_D(WebSocketModule);
+
+    qDebug() << "WebSocketModule::connect with args: url: " << url << " socketId: " << socketId;
```
vkjr: Quite a lot of qDebug() calls here and there. Maybe we can wrap them with `#ifdef DEBUG`?

MaxRis: What if to check for RCT_DEV? RCT_DEV is enabled in Debug only.

MaxRis: It seems pretty a lot interesting things happen via WebSocket. Should be good to know that something happens there.

Issue #143, opened 2017-11-17
Nested <Text> doesn't work after merging with React Native 0.50


vkjr:
Non-working example (can be found in RNTester, TextExample)

```
          <Text selectable={true}>
            This text is <Text style={{fontWeight: 'bold'}}>selectable</Text> if you click-and-hold.
          </Text>
```

metellius:
Doesn't work at all, or even crashes reactqt as in my case?

vkjr:
@metellius, it crashes

PR #160, opened 2017-12-15
Fixed crash on nested texts

vkjr:
Now text component can be with or without associated flexbox node. Only topmost text node among nested ones have a flexbox node.

### DIFF
--- a/ReactQt/runtime/src/qml/ReactText.qml
+++ b/ReactQt/runtime/src/qml/ReactText.qml
@@ -19,7 +19,7 @@ Text {
     property color p_textDecorationColor
     property string p_writingDirection
     property int p_numberOfLines: 1000000
-    property var flexbox: React.Flexbox {control: textRoot}
+    property var flexbox: null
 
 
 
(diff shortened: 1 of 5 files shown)


vkjr:
Related to #143

ReactQt/runtime/src/qml/ReactText.qml
```diff
@@ -65,6 +65,20 @@ Text {
     onParentChanged: updateHtmlText()
 
     function updateMeasureFunction() {
+
+        //Only topmost text item in a set of nested ones can have a flexbox node.
+        if(textIsTopInBlock) {
+            textRoot.flexbox = Qt.createQmlObject('import React 0.1 as React; React.Flexbox {control: textRoot}',
```
MaxRis: If I'm not wrong, might be less confusing to put `parent` instead of `textRoot` inside new QML object code.

vkjr: @MaxRis, I checked, that doesn't work because Flexbox is not a QQuickItem, it is a QObject* and it doesn't have `parent` property.

MaxRis: @vkjr thank you!

PR #163, opened 2017-12-20
Set react version to 16.0.0

MaxRis:


MaxRis:
@vkjr going to merge react version update into feature/run-status-im branch.
In next steps will try to use primary react-native-qt branch from status-react develop-desktop.

PR #164, opened 2017-12-20
fixing README typo

alex-min:
Just fixing a small README typo


vkjr:
@alex-min, thank you.

PR #176, opened 2018-01-03
Extend Bridge to send promise callbacks with args in json map

MaxRis:


ReactQt/runtime/src/bridge.cpp
```diff
@@ -187,11 +187,18 @@ void Bridge::enqueueJSCall(const QString& module, const QString& method, const Q
...
 
 void Bridge::invokePromiseCallback(double callbackCode, const QVariantList& args) {
+    invokePromiseCallbackInternal(QVariantList{callbackCode, args});
+}
+
+void Bridge::invokePromiseCallback(double callbackCode, const QVariantMap& args) {
+    invokePromiseCallbackInternal(QVariantList{callbackCode, args});
+}
+
+void Bridge::invokePromiseCallbackInternal(const QVariantList& args) {
```
vkjr: I would suggest to use another name that describes what function actually does, f.ex. `invokePromiseViaExecutor()`. Or something like this :)

MaxRis: @vkjr thanks! Was not able to find good enough naming for common method. "ViaExecutor" should work, but it seems we don't have others ways how to run JS code. Also since the method is private, the name of it is essential only in scope of Bridge's internals.

PR #181, opened 2018-01-10
Do not size ScrollView to its parent

MaxRis:


ReactQt/runtime/src/qml/ReactScrollView.qml
```diff
@@ -9,7 +9,6 @@ Flickable {
     property var flexbox: React.Flexbox {control: scrollViewRoot}
 
     clip: true
-    anchors.fill: parent
```
MaxRis: @vkjr not sure why this was added initially. Maybe detail of old implementation without flex layout? ScrollViewExample is rendered fine without these anchors and scrolling area issues in status-react's chat screen are solved!

vkjr: @MaxRis, not sure but it could be that I added this line to launch RNTester (before correct flex layout implementation).

PR #183, opened 2018-01-25
Generate onLayout View event on layout changed

MaxRis:


ReactQt/runtime/src/layout/flexbox.cpp
```diff
@@ -155,10 +157,23 @@ void FlexboxPrivate::updatePropertiesForControlsTree(YGNodeRef node) {
 void FlexboxPrivate::updatePropertiesForControl(YGNodeRef node) {
     auto qmlControl = static_cast<FlexboxPrivate*>(YGNodeGetContext(node))->m_control;
 
+    bool emitLayoutUpdated = false;
```
vkjr: If this improves readability, we can write this statement as:
`bool emitLayoutUpdated = YGNodeLayoutGetLeft(node) != qmlControl->x() || ...`

MaxRis: thanks!

PR #186, opened 2018-01-29
Improvement/text background color

vkjr:
backgroundColor property support added to <Text> component


vkjr:
@MaxRis, sorry, created this PR on the base of previous.

PR #187, opened 2018-01-29
Emit layout changed native events

MaxRis:
- Emit native RN "onContentSizeChange" for TextInput
- Emit native RN "onLayout" for Text and TextInput. Refactored emitting of "onLayout" for View.

### DIFF
--- a/ReactQt/runtime/src/componentmanagers/textinputmanager.cpp
+++ b/ReactQt/runtime/src/componentmanagers/textinputmanager.cpp
@@ -31,6 +31,7 @@ const QString EVENT_ON_END_EDITING = "onEndEditing";
 const QString EVENT_ON_FOCUS = "onFocus";
 const QString EVENT_ON_BLUR = "onBlur";
 const QString EVENT_ON_KEY_PRESS = "onKeyPress";
+const QString EVENT_ON_CONTENT_SIZE_CHANGE = "onContentSizeChange";
 
 class TextInputManagerPrivate {};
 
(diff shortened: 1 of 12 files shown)


vkjr:
@MaxRis, did I understood correctly that currently steps for updating layout are:

1. RootView invoke flexbox->recalculateLayout
2. Flexbox of RootView calls updatePropertiesForControl() on every control in a tree
3. updatePropertiesForControl() checks control sizes and invokes layoutUpdated() signal
4. layoutUpdate() signal leads to sendLayoutUpdatedToJs() invokation

If so, then we will get layoutUpdate notification when resize root view, but won't get if size of the control changed programmatically.

Maybe we should use another order of calls? For example:
1. ViewManager when creates control, subscribes to its size changes
2. If size changed, we invoke sendLayoutUpdatedToJs() and also request flexbox->recalculateLayout() on a root view.

This way we can avoid adding onLayout() signal to every qml control, they all have managers and all managers derived from ViewManager.

What do you think?

MaxRis:
@vkjr this sounds reasonable, but, I believe, some additional investigation required, which I haven't done at this stage.
ViewPropTypes.js has following definition of onLayout prop
```
  /**
   * Invoked on mount and layout changes with:
   *
   * `{nativeEvent: { layout: {x, y, width, height}}}`
   *
   * This event is fired immediately once the layout has been calculated, but
   * the new layout may not yet be reflected on the screen at the time the
   * event is received, especially if a layout animation is in progress.
   */
  onLayout: PropTypes.func,
```

And it seems, onLayout native event is really triggered for iOS before ui elements get updated.
RCTUIManager.m method - (RCTViewManagerUIBlock)uiBlockWithLayoutUpdateForRootView:(RCTRootShadowView *)rootShadowView:

1. RCTRootShadowView triggers YGNodeCalculateLayout to recalculate layout and returns set of objects NSMutableSet<RCTShadowView *> (seems doesn't change UI yet)
2. onLayout native event is triggered
3. some layoutAnimation calculation is done to update ui elements (in my understanding). `[view reactSetFrame:frame];` is called.

My understanding is that there exists some important reasons why the event is triggered before ui elements properties are changed.

MaxRis:
@vkjr it seems it really doesn't make sense to touch any QML files in chain of onLayout event generation. It should have more sense to trigger ViewManager directly from flexbox and pass new layout values with view's tag. Will try to update changes. Thank you

MaxRis:
@vkjr have updated flexbox to send onLayout events directly via ViewManager. But to keep the implementation identical what iOS has and, seems, expected for onLayout event, it's emitted just after the layout is recalculated and before ui elements are updated.

vkjr:
@MaxRis, thank you for explaining onLayout behavior. Before this I thought that it is called after size changed :)

MaxRis:
@vkjr I will rename `ViewManager::sendLayoutUpdatedToJs` to `ViewManager::sendOnLayoutToJs` to avoid confusion, thank you!

Issue #192, opened 2018-02-09
Package application for distribution


MaxRis:
- `react-native package` command
- Use LocalServerConnection connection to communicate with node process
- Embed js sources into resources
- Build application in Release

cc: @vkjr

PR #200, opened 2018-02-21
refs #192 - Use LocalServerConnection by default

MaxRis:
- LocalServerConnection (pipe for node process) is used by default
- `react-native run-desktop` updated to use `RemoteServerConnection` by default, if anything else is not specified
- Redirect console.info to console.error stream if process pipe is used for JS server
- Enable high dpi displays support for users apps by setting  QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);

------

package.json of custom app is extended with support of "desktopJSBundlePath" property to put path to JS bundle which will be embedded into final binary as resource and loaded from there (instead of querying rn-packager)

### DIFF
--- a/ReactQt/runtime/src/sourcecode.cpp
+++ b/ReactQt/runtime/src/sourcecode.cpp
@@ -13,6 +13,7 @@
 
 #include <memory>
 
+#include <QFile>
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
 #include <QNetworkRequest>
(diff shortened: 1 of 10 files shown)


ReactQt/runtime/src/sourcecode.cpp
```diff
@@ -86,24 +88,38 @@ void SourceCode::setRetryCount(int retryCount) {
...
-            if (d->retryAttempts < d->retryCount) {
-                d->retryAttempts++;
-                d->retryTimeout *= 2;
-                QTimer::singleShot(d->retryTimeout, [=] { loadSource(nam); });
-            } else {
-                qCritical() << __PRETTY_FUNCTION__ << ": Error while loading source" << reply->errorString();
-                Q_EMIT loadFailed();
+    const QString JS_BUNDLE_RESOURCES_PATH = QStringLiteral(":/index.desktop.bundle");
+    if (QFile::exists(JS_BUNDLE_RESOURCES_PATH)) {
+        QTimer::singleShot(1000, [=]() {
```
MaxRis: @vkjr here happens loading of embedded JS bundle from binary's resources. Delay is a workaround to make sure that node server is started fully (creation of sandbox and precreation of Realm instances requires some time). Proper solution instead of delay, seems, might be receiving of some special message from node server, that all setup is done and we can load custom app JS code for execution.

vkjr:
@MaxRis, do I understand correctly that all error messages that we were receiving in ./ubuntu-server.js output now will be shown in application output?

local-cli/desktop/runDesktop.js
```diff
@@ -71,8 +71,11 @@ function actuallyRun(args, reject) {
         appArgs.push('--host=' + args['host']);
       if (args['port'])
         appArgs.push('--port=' + args['port']);
-      if (args['executor'])
+      if (args['executor']) {
         appArgs.push('--executor=' + args['executor']);
+      } else {
+        appArgs.push('--executor=RemoteServerConnection');
```
MaxRis: > do I understand correctly that all error messages that we were receiving in ./ubuntu-server.js output now will be shown in application output?
@vkjr , if you have start application as `react-native run-desktop` it will replace default connection(piped process) with RemoteServerConnection(as it referenced in the source code above). But, yes, if we start application directly, for example, from root directory of project it will be like: "./desktop/bin/ProjectBinary" ... it will use started with piped node process and all JS output will be redirected to the native app process output.

MaxRis:
As summary for above changes:

- It's required to pre-create JS bundle source code file which will be embedded as resource via package.json Commands:
```
To get bundle from inside react-native-desktop directory:
curl http://localhost:8081/index.desktop.bundle?platform=desktop -o index.desktop.bundle

To get bundle from react-native app:

react-native bundle --entry-file index.desktop.js --bundle-output ./StatusIM.jsbundle --dev false --platform desktop
```
- ubuntu_server.py still discovered as separate file in working directory (TODO: Embed it as resource if possible)
- ubuntu_server.py for StatusIM loads Realm on start of node process. Not sure yet how to move forward with application distribution with such dependency on Realm package from node_modules folder.
- There is existing some issue with resources loading (by Image component, for example).It seems images URLs expects rn-packager running, but it's not. Requites investigation.

local-cli/desktop/runDesktop.js
vkjr: @MaxRis, thanks, got it :)

PR #202, opened 2018-03-01
Master

ndpmurphy:
Transpose did not work.

## Test Plan
Examples/Transpose contains a demo that shows Views and Images transposed in various ways.
It shows that 2d transforms now work. Transforms with perspective seem to get clipped in some odd way and do not match the appearance when run on IOS. Also the transformed images have random background colours. So these are 2 things that could be fixed later.


vkjr:
@ndpmurphy, could you please check this pull request. I think there is a mistake here. You are suggesting to commit changes from your "master" branch, but it doesn't contain any commits with transform fixes. 

`master` branch in `react-native-desktop` is an old code that was a base for our development. All the latest code exists in `react-native-qt` branch.
To create a pull-request with new code for transformation you can do the following:
- fork `react-native-desktop`
- create own development branch based on `react-native-qt` branch, for example `feature-fix-transformation`
- add code for fixing transformation to created branch
- create pull request and specify your `ndpmurphy/feature-fix-transformation` branch as a souce.

Thanks.

ndpmurphy:
Sorry for making  a mess of that first pull request. I have a new one at https://github.com/status-im/react-native-desktop/pull/203.  It is failing some of the CI tests whcih confuses me  because they pass locally on my own machine. So I will continue to investigate that.

Issue #196, opened 2018-02-13
transform does not work


ndpmurphy:
In the examples directory the Image example does a transpose of a png. The image is the word google. If you run it on IOS then the word google appears once and is disappearing into the distance with a 'perspective' and rotation applied as a transpose.

On Ubuntu if I run the same example I see several copies of the bitmap with no rotation or perspective applied. 

I expected this to work since there was a commit from  @justincan  back in 2016 to implement transform. It was commit 0e25614c1dc023b935221abf12d7aba919ba583f. Maybe something broke it afterwards.

### Is this a bug report?
Yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?
YES

### Environment
Scanning folders for symlinks in /home/osboxes/AwesomeProject/node_modules (19ms)

Environment:
  OS: Linux 4.13
  Node: 8.9.4
  Yarn: Not Found
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Not Found

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: Not Found

I am not sure why it says react-native not found - I assume it is because react native is not in package.json because I am at the root of the react-native  repo. Git hash  058c0f424fcc38029b8e6cbd82907779edc0e6d6 on branch  react-native-qt

### Steps to Reproduce
1. Install the status-im/react-native-desktop repo
2. Build the QT project
3. Run the image example

### Expected Behavior

I expect to see the word google with perspective and rotation applied ( as per IOS)
![image](https://user-images.githubusercontent.com/5898278/36143344-bfb6a33a-10a2-11e8-850a-36b49c2c4cda.png)

### Actual Behavior

See attached screenshot, of Google png ppearing multiple times
![image](https://user-images.githubusercontent.com/5898278/36143292-9f325ca8-10a2-11e8-936f-1bb73d558889.png)

I also reproduced this problem in one of my own programs where I tried to apply a transform with translateX to a view. It was a simpler case than the ImageApp example, but it also failed. In that case the view simply stayed where the layout engine put it, but the transform was  ignored. Again, onIOS it worked fine.

ndpmurphy:
[...]

PR #203, opened 2018-03-02
Fix Transpose ( issue #196 ), and add an example app.

ndpmurphy:
Transpose did not work.

## Test Plan

Examples/Transpose contains a demo that shows Views and Images transposed in various ways.
It shows that 2d transforms now work. Transforms with perspective seem to get clipped in some odd way and do not match the appearance when run on IOS. Also the transformed images have random background colours. So these are 2 things that could be fixed later.


ReactQt/runtime/src/reactitem.cpp
```diff
@@ -393,6 +412,17 @@ double ReactItem::shadowRadius() const {
 
 void ReactItem::setShadowRadius(double shadowRadius) {}
 
+QVector<float> ReactItem::transform() const {
+    return QVector<float>{};
```
vkjr: @ndpmurphy, please return here the same value that was passed to `setTransform()` function.
We can maintain `QVector<float> _transform` member, save value passed in `setTransform` to it and return it from `transform()`

Examples/Transform/index.js
```diff
@@ -0,0 +1,167 @@
+
+// Display views and an image with various transforms applied
```
vkjr: @ndpmurphy, please, use the same naming for example js files as in another examples, `TransformApp.js` in this case.
Also example should be added to `Examples/CMakeLists.txt`

ndpmurphy: I have made this change, but can you tell me what is the standard way of running the Examples. I have been manually editing the main.qml.in file to set the source file and moduleName. This works, but I am sure the CMakefiles must provide a better way to launch an example from with QtCreator or from the commandline.

vkjr: @ndpmurphy, here is how you can setup "Run" section in QtCreator to launch examples using `qmlscene`:

screen shot 2018-03-20 at 8 59 39 am

ndpmurphy: But there is no Image.qml file in the Image directory. Was this supposed to be generated by CMakefile ?

vkjr: @ndpmurphy, this command invoked in `%{buildDir}`, which is build directory. You will get Image.qml there after build.

In `Examples/Image/CMakeLists.txt` you can find following code:
```
configure_file(
  ../Example.qml.in
  ${CMAKE_CURRENT_BINARY_DIR}/${EXAMPLE_NAME}.qml
)
```

It is responsible for generating `Image.qml`

ndpmurphy: That does not work for me. If I open the CMakefiel at the top level, then I can not  get it to build. I have been opening the CMakefile in the ReactQT directory and that works fine for most things, but it does not generate the qml files in Examples. When opening the project, is the CMAkefile at the top levelthe right file to open or is there another project file ?

ndpmurphy: Oh - ignore that I see my mistake now

vkjr: @ndpmurphy, were you able to run the example apps the way I mentioned above?
I usually open root CMakeLists.txt in QtCreator. Then you will be able to build both, ReactQt and Examples projects.

Issue #215, opened 2018-05-03
JavaScriptCore JS engine support


MaxRis:
It should be possible optionally to replace Node.JS with JavaScriptCore

PR #216, opened 2018-05-03
Optional JavaScriptCore JS engine support #215

MaxRis:
Optional JavaScriptCore engine support on Linux:

- CMAKE option -DJAVASCRIPTCORE_ENABLED should be specified to enable build of JavaScriptCore and Facebook folly libs + react-native-desktop sources related to JavaScriptCore. (**ATTENTION**: Building of JavaScriptCore requires at least 8GB RAM, took 10+ GB of disk space and downloads 8GB of data from WebKit GitHub repository, lasts 15+ minutes)
- JavaScriptCoreExecutor class introduced to encapsulate all internal work related to JavaScriptCore

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,16 @@ project(react-native-desktop)
 set(PROJECT_NAME "react-native-desktop")
 set(PROJECT_VERSION "0.0.1")
 
+option(JAVASCRIPTCORE_ENABLED
+    "Build with JavaScriptCore enabled"
+    OFF)
+
+if(JAVASCRIPTCORE_ENABLED)
+  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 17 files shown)


vkjr:
@MaxRis, did I understand correctly that -DJAVASCRIPTCORE_ENABLED should be added in desktop/build.sh cmake invokation line?

vkjr:
@MaxRis, Does react-native use JavaScriptCore as a default engine for ios and android apps?

MaxRis:
@vkjr it should work, but at this time, it should be better just to modify corresponding option from OFF to ON, if you would like to try it https://github.com/status-im/react-native-desktop/pull/216/files#diff-af3b638bc2a3e6c650974192a53c7291R18
P.S. Next time you decide to change the option value back, you will need manually to remove CMake build cache.

MaxRis:
@vkjr yes, on Mac OS and iOS there is available JavaScriptCore framework. On Android, not sure, but, probably, it's part of Android SDK.

PR #227, opened 2018-05-17
Fixed key string in TextInput onKeyPress event

vkjr:
<TextInput> control now returns correct strings ("Enter" or "Backspace") for Enter and Backspace keys.


vkjr:
TextInput control now returns correct strings ("Enter" or "Backspace") for Enter and Backspace keys.

PR #234, opened 2018-05-25
Add cli to npm package; Use patch-package to patch metro bundler; Rem…

MaxRis:
- react-native-cli folder is added into package.json and deployed on `npm install`. I've found that the easiest way to install react-native-cli on Jenkins CI builds. Should be useful in other cases because of this.
- metro packager is patched automatically on `npm install` with help of tool `patch-package`. Tool provides easy way to create new patches on top of any used npm package.
- Deleted not used template files


MaxRis:
These changes are replaced with PR https://github.com/status-im/react-native-desktop/pull/235
react-native-cli package is not added to react-native-desktop node package finally, since such adding introduces JS files conflicts.

PR #263, opened 2018-06-17
Merge/react native v.0.55.4

MaxRis:
- All merged commits history is converted into single commit to not brake our own commits history
- patch-package patches updated for used newer version of metro bundler packager


vkjr:
@MaxRis, I'm trying to create simple project.
Changed line in `react-native-cli/index.js` to point to your repository, installed that react-native-cli globally, then as usually `react-native init TestMerge`, `cd TestMerge` and on invoking `react-native desktop` I'm getting "unrecognized command`

Looks like `genDesktop` command is missed in `local-cli/commands.js`.
Could you please check? Did I miss something?

PR #265, opened 2018-06-19
Merge/react native v0.55.4

MaxRis:
- All merged commits history is converted into single commit to not brake our own commits history
- patch-package patches updated for used newer version of metro bundler packager


MaxRis:
Hey @vkjr , I'm sorry, have created previous PR for wrong branch.
Current branch is tested with new app creation and also already tested with status-react desktop. Will create special PR to update react-native version in status-react as well after merge of these changes. Thanks

Issue #268, opened 2018-06-21
Command `run-desktop` unrecognized. (after following Getting Started)


msh2050:
### Is this a bug report?

yes

### Have you read the [Contributing Guidelines](https://facebook.github.io/react-native/docs/contributing.html)?

yse

### Environment

Environment:
  OS: Windows 10
  Node: 8.11.3
  Yarn: 1.7.0
  npm: 5.5.1
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Version  3.1.0.0 AI-173.4720617

Packages: (wanted => installed)
  react: 16.3.1 => 16.3.1
  react-native: https://github.com/status-im/react-native-desktop#master => 0.55.4

### Steps to Reproduce

(Write your steps here:)

1. Just fallow the steps in Getting Started(Install Prerequisites , Install adjusted react-native-cli and Create new App)

### Expected Behavior

I just expect a hello world screen will run  
### Actual Behavior

after give the command (  >react-native desktop)
I get the following error (

> exec error: Error: Command failed: pwd && patch --verbose -d ./node_modules/metro/src/JSTransformer -i ../../../react-native/increase_bundler_memory_usage.patch
> 'pwd' is not recognized as an internal or external command,
> operable program or batch file.

) 
then:

>  Command `run-desktop` unrecognized. Make sure that you have run `npm install` and that you are inside a react-native project.

please find the below attached cmd script from start to end. note how the command "info" and  help changing after the command (  >react-native desktop)

[cmdscript.txt](https://github.com/status-im/react-native-desktop/files/2124362/cmdscript.txt)
### Reproducible Demo

(Paste the link to an example project and exact instructions to reproduce the issue.)

madhavarshney:
> removed 1212 packages and updated 18 packages in 15.2s

Seems like all the packages got removed when running `react-native desktop`.  This is probably due to `react-native init` using yarn to bootstrap the project and then `react-native desktop` running an install command with npm, which removes all the other packages.  Can you run `yarn`/`yarn install` in the project and try again?

vkjr:
@msh2050, thank you for reporting.

>'pwd' is not recognized as an internal or external command,
'pwd' is not a Windows command. 
I think`react-native-desktop/local-cli/desktop/genDesktop.js` should be modified to make patching cross-platform to resolve this issue.

msh2050:
I get this error:
```
Command failed: build.bat -g "MinGW Makefiles"
[...]

PR #274, opened 2018-06-26
add npm install to list of steps for new app creation

rcullito:
add npm install to list of steps for new app creation

### DIFF
--- a/docs/ReactQt/CreateNewApp.md
+++ b/docs/ReactQt/CreateNewApp.md
@@ -7,6 +7,7 @@ In order to setup the RN-desktop project, execute these terminal commands:
 ```
 ~$ react-native init DesktopSampleApp && cd DesktopSampleApp
 ~/DesktopSampleApp$ react-native desktop
+~/DesktopSampleApp$ npm install
 
 ```
 If you're using macOS, run these commands in separate shells:


madhavarshney:
@rcullito - My two cents: if I have not misinterpreted the problem, this should not be needed after running the specified steps.  `react-native init` automatically adds and installs the required dependencies with yarn, if yarn is present, or else npm.  `react-native desktop` installs the corresponding version of react here:

https://github.com/status-im/react-native-desktop/blob/a0acfc48d2b34902ca35d4c7279828d6082252ca/local-cli/generator/index.js#L98-L115

This causes a problem if `react-native init` installed the dependencies with yarn, as now npm removes all the other dependencies due to yarnpkg/yarn#5240 and https://github.com/npm/npm/issues/17379#issuecomment-345042377, demonstrated by https://github.com/status-im/react-native-desktop/issues/268#issuecomment-399158695.  A better fix would be to remove that block of code as react-native-desktop's cli is used to initialize a project and installs the correct version of react.

If you're trying to solve a different problem, then nevermind.

rcullito:
Thanks very much for this @madhavarshney ! I was unfamiliar with the yarn/npm incompatibilities, so much appreciated on bringing it up and linking to the relevant issues.  

So to recap: this lingering `npm install` is wiping out the yarn dependency tree? And we can remove this block of code because we already have the correct version of react in place anyway? :pray:

PR #271, opened 2018-06-22
Cleanup left-overs from previous merges

madhavarshney:
Remove left-over files from previous react-native merges that are no longer needed.

### Motivation

Cleanup...

### Test Plan

N/A. This should (hopefully) fix CI/CD build break.

#### Related PRs
N/A

#### Release Notes
N/A

PR #276, opened 2018-06-27
documentation fixes

rcullito:
fix some links that weren't working as well as take out the `npm install` addition so that it isn't misleading.

For what it's worth, I experimented with removing the install block within `generator/index.js` but after re-installing `react-native-cli` I still received the yarn/npm issue on a newly created app. 

Would an additional command have to be run @MaxRis for any local changes to the cli tool in order for them to be bundled up and captured as part of a subsequent npm install? Or I could have simply been missing something. Let me know what you all think is sensible. Thanks! cc @madhavarshney

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,16 +28,16 @@ To access `In-App Developer Menu` ~~start shaking your laptop/PC~~ press CTRL+R.
 
 ### 3rd-party react-native modules support
 Pure JavaScript react-native modules can be used with react-native-desktop without limitations.  
-Creation of custom react-native `native` modules or `adopting of available native modules` is supported and described in following [guideline](docs/ReactQt/NativeModulesSupport.md).  
+Creation of custom react-native `native` modules or `adopting of available native modules` is supported and described in following [guideline](docs/NativeModulesSupport.md).
 
-[List of supported or partially supported native modules](docs/ReactQt/SupportedNativeModulesList.md).
+[List of supported or partially supported native modules](docs/SupportedNativeModulesList.md).
 
 ## Documentation
 
-- [How react-native-desktop works internally](docs/ReactQt/HowRNDesktopAppWorks.md)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,9 +16,9 @@ Recommended Qt framework of version 5.9.1 LTS.
 
 ## Getting Started
 
-- [Install Prerequisites](docs/ReactQt/InstallPrerequisites.md)  
-- [Install adjusted react-native-cli](docs/ReactQt/InstallUpdatedReactNativeCLI.md)  
-- [Create new App](docs/ReactQt/CreateNewApp.md)  
+- [Install Prerequisites](docs/InstallPrerequisites.md)
```
madhavarshney: I thought I had done this in #271. Sorry for the confusion. 😕

madhavarshney: Can you fix all of the links to the docs? `docs/ReactQt/*` -> `docs/*`

madhavarshney:
@rcullito - If I understood correctly, you removed that block of code from `local-cli/generator/index.js` and then re-installed `react-native-cli`.  This wouldn't have applied the fix, however, since `react-native-cli` installs `react-native-desktop` from _GitHub_ and then invokes it's local-cli from there.

README.md
rcullito: no worries! super small change. and good call, just pushed the rest.

rcullito:
ah thanks @madhavarshney ! will use a symlink then

madhavarshney:
@rcullito - I haven't had very good experience with symlinks.  You should commit your changes, push them to a branch on your fork, modify this line to point to your fork and branch, and then reinstall react-native-cli.
https://github.com/status-im/react-native-desktop/blob/492eb0aa68b47b107de2799ba3fe436cb617f237/react-native-cli/index.js#L259

Issue #253, opened 2018-06-13
UIManager::manageChildren arrays reconciliation support


MaxRis:
`UIManager::manageChildren` implementation introduces issues on some scenarios.

Possibly, it happens because input argument `addAtIndices` may contain any range of values to add new element at, but current implementation, seems, adds new items only to the end of parent containers list (QQuickItem API to insert new child at specific position seems is not available ? ):

```
child->setParentItem(container);
```

JS source sample with issues:

```
import React, { Component } from 'react';
import {
 AppRegistry,
 Alert,
 Text,
 TouchableHighlight,
 View,
 TextInput,
 Button
} from 'react-native';



class ItemList extends Component {
 constructor(props) {
   super(props);
   this.state = {longList: true};
   this.onPress = this.onPress.bind(this);
 }

 onPress() {
   this.setState({longList: !this.state.longList});
 }

 onButtonPress() {
  this.setState({longList: !this.state.longList});
 }

 render() {
   let items = this.state.longList ? ["Long11", "Long22", "Long23"] : ["Long22", "Long23"];
   return (
     <View onPress={this.onPress}>

       <View>
       {items.map(function(name, index) {
         return <Button key={name} title={name} > </Button>;
       })}
       </View>
     </View>
   )
 }
}

class rn_list extends Component {
 render() {
   return (
     <ItemList/>
   );
 }
}

AppRegistry.registerComponent('rn_list', () => rn_list);
``

The issue doesn't happen if in sample above to update line with arrays values to the following:

```
let items = this.state.longList ? ["Long11", "Long22", "Long23"] : ["Long11", "Long22"];
```

Nice to implement some automated tests to verify work results of UIManager::manageChildren

PR #279, opened 2018-06-30
Update CI

madhavarshney:
### Motivation

A few updates and changes:

- Updated the base docker image from Ubuntu Trusty 14.04 to Ubuntu Xenial 16.04.
- Added a step to update NPM to v6.
- Updated package-lock.json.
- Removed CircleCI caching of `node_modules` in favor of [`npm ci`](https://docs.npmjs.com/cli/ci) + CircleCI caching of NPM cache.
- Removed the [cmake PPA](https://launchpad.net/~george-edison55/+archive/ubuntu/cmake-3.x) in favor of the official [cmake package](https://packages.ubuntu.com/xenial/cmake).

_Note: This PR does not upgrade Ubuntu to Bionic (18.04) because a [Qt 5.9 LTS PPA](https://launchpad.net/~beineri) is not available for that version.  If using Qt 5.11 is fine, then that can be done as well._

### Test Plan

N/A.

### Related PRs

N/A.

### Release Notes

This PR upgrades the npm lockfile version to npm@6.

PR #280, opened 2018-07-02
Insert child QMLItem at specific position #253

MaxRis:
- `UIManager::manageChildren` should insert childs at specific position in parent container, but not just add to the end as it does currently
- Unit tests to validate calling of `UIManager::manageChildren` in different contexts
- CI updates to fail sh commands if part of command fails

Closes https://github.com/status-im/react-native-desktop/issues/253

P.S. Just discovered that there are mysterious [QQuickItem::stackBefore](http://doc.qt.io/qt-5/qquickitem.html#stackBefore) and stackAfter to reorder the childs, but so far it's totally unclear how to use them properly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,8 +36,8 @@ jobs:
           sudo add-apt-repository -y ppa:beineri/opt-qt591-trusty
           sudo add-apt-repository -y ppa:george-edison55/cmake-3.x
           sudo apt-get -qy update
-          sudo apt-get install -y dbus xvfb curl cmake mesa-common-dev \
-              qt59base qt59graphicaleffects qt59quickcontrols2 qt59declarative qt59websockets
+          sudo apt-get install -y --force-yes dbus xvfb curl cmake mesa-common-dev \
+              qt59base qt59graphicaleffects qt59quickcontrols2 qt59declarative qt59websockets qt59svg
           echo "/opt/qt59/bin/qt59-env.sh" >> ~/.circlerc
 
       # Install node
@@ -54,16 +54,16 @@ jobs:
 
       # Build react-native-desktop in Debug (RCT_DEV enabled)
       - run: |
-          mkdir build_debug
-          pushd build_debug
-          cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j4
+          mkdir build_debug && \
+          pushd build_debug && \
+          cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j4 && \
           popd
 
       # Build react-native-desktop in Release
       - run: |
-          mkdir build
-          pushd build
-          cmake -DCMAKE_BUILD_TYPE=Release .. && make -j4
+          mkdir build && \
+          pushd build && \
+          cmake -DCMAKE_BUILD_TYPE=Release .. && make -j4 && \
           popd
 
       # Download and cache dependencies
@@ -71,7 +71,7 @@ jobs:
           key: dependency-cache-{{ checksum "package.json" }}
 
       - run: |
-          rm -rf ./node_modules/metro
+          rm -rf ./node_modules/metro && \
           npm install
 
       # Apply metro patch to add desktop platform
@@ -97,6 +97,6 @@ jobs:
 
       # run unit tests
       - run: |
-          pushd build
-          xvfb-run --server-args="-screen 0 1024x768x24" ctest -VV
+          pushd build && \
+          xvfb-run --server-args="-screen 0 1024x768x24" ctest -VV && \
           popd
--- a/ReactQt/tests/test-array-reconciliation.cpp
+++ b/ReactQt/tests/test-array-reconciliation.cpp
@@ -1,0 +1,138 @@
+/**
+ * Copyright (c) 2017-present, Status Research and Development GmbH.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include <QSignalSpy>
+#include <QTest>
+#include <QTimer>
+#include <QtQuick/QQuickView>
+
+#include "reacttestcase.h"
+#include "redbox.h"
+
+class TestArrayReconciliation : public ReactTestCase {
+    Q_OBJECT
+
+private slots:
+    virtual void initTestCase() override;
+    CLEANUP_TEST_CASE_DEFAULT(ReactTestCase)
+
+    void testComponentsArrayFirstElementInsert();
+    void testComponentsArrayLastElementDelete();
+
+private:
+    QTimer* clickDelayTimer = nullptr;
+    QQuickItem* topView = nullptr;
+    const int INITIAL_ITEMS_COUNT = 3;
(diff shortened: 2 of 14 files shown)


.circleci/config.yml
```diff
@@ -36,8 +36,8 @@ jobs:
           sudo add-apt-repository -y ppa:beineri/opt-qt591-trusty
           sudo add-apt-repository -y ppa:george-edison55/cmake-3.x
           sudo apt-get -qy update
-          sudo apt-get install -y dbus xvfb curl cmake mesa-common-dev \
-              qt59base qt59graphicaleffects qt59quickcontrols2 qt59declarative qt59websockets
+          sudo apt-get install -y --force-yes dbus xvfb curl cmake mesa-common-dev \
```
madhavarshney: From [the apt-get linux man page](https://linux.die.net/man/8/apt-get):

> Force yes. This is a dangerous option that will cause apt-get to continue without prompting if it is doing something potentially harmful. It should not be used except in very special situations. Using --force-yes can potentially destroy your system!

Is this done to circumvent some problems after adding `qt59svg`?

.circleci/config.yml
```diff
@@ -54,24 +54,24 @@ jobs:
 
       # Build react-native-desktop in Debug (RCT_DEV enabled)
       - run: |
-          mkdir build_debug
-          pushd build_debug
-          cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j4
+          mkdir build_debug && \
+          pushd build_debug && \
+          cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j4 && \
```
madhavarshney: Not sure what this is supposed to change...

MaxRis: @madhavarshney without making it as one command the whole `sh` CircleCI command doesn't fail if some inner command exits with error code. Because of that all previous ci runs we had compilation with CMake failed, but it was displayed as passed.

.circleci/config.yml
MaxRis: @madhavarshney this was done after failures in following ci run https://circleci.com/gh/MaxRis/react-native-desktop/341
The errors were:

```
Reading package lists...
W: GPG error: http://ppa.launchpad.net trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 10C56D0DE9977759
W: GPG error: http://ppa.launchpad.net trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 084ECFC5828AB726
...
WARNING: The following packages cannot be authenticated!
  cmake-data cmake qt59base qt59xmlpatterns qt59declarative
  qt59graphicaleffects qt59quickcontrols2 qt59svg qt59websockets
E: There are problems and -y was used without --force-yes
Exited with code 100
```
Maybe that was some temp issue with packages.
Probably, better to left this flag as is, since nothing very bad can happen on CircleCI's docker image, hopefully, and we don't want to experience such failures in the future :)

ReactQt/tests/test-array-reconciliation.cpp
```diff
@@ -0,0 +1,163 @@
...
+    clickDelayTimer->start();
+    waitAndVerifyCondition([=]() { return !clickDelayTimer->isActive(); }, "Timer timeout was not triggered");
+    waitAndVerifyCondition([=]() { return topView->childItems().size() == CLICKED_ITEMS_COUNT; },
+                           "Wrong array items count after 1st click");
+
+    topView = topJSComponent();
+    textInput = topView->childItems().at(0);
+    clickItem(textInput);
+
+    clickDelayTimer->start();
```
vkjr: These 3 lines of code (starting timer and 2 waitAndVerify calls) can be extracted into a separate function to reduce code duplication in the file.

MaxRis: Yep, sure, thanks

.circleci/config.yml
madhavarshney: Doesn't seem to be true: https://circleci.com/gh/status-im/react-native-desktop/566.  I think cmake doesn't exit with a non-zero exit code on configuring errors.

.circleci/config.yml
madhavarshney: I think it was a temp issue with packages.  For some reason GPG could not fetch the proper keys. 
 Relevant section of GPG errors:

```
gpg: keyring `/tmp/tmpjovwmzt3/secring.gpg' created
gpg: keyring `/tmp/tmpjovwmzt3/pubring.gpg' created
gpg: requesting key E9977759 from hkp server keyserver.ubuntu.com
gpgkeys: key C65D51784EDC19A871DBDBB710C56D0DE9977759 can't be retrieved
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
gpg: keyring `/tmp/tmpvtm636tq/secring.gpg' created
gpg: keyring `/tmp/tmpvtm636tq/pubring.gpg' created
gpg: requesting key 828AB726 from hkp server keyserver.ubuntu.com
gpgkeys: key B118CD3C377DF930EDD06C67084ECFC5828AB726 can't be retrieved
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
```

Either way, I think the best idea would be to create a custom docker image with CMake, Qt, and maybe Node.js preinstalled, as that's the [recommended way to do persistent apt-get installs](https://discuss.circleci.com/t/caching-apt-get-packages/9076/3) and would make sure any such issues don't appear.

.circleci/config.yml
MaxRis: Indeed, cmake does according to following logs:
https://circleci.com/gh/MaxRis/react-native-desktop/335
https://circleci.com/gh/MaxRis/react-native-desktop/336
Please compare the output for cmake commands between 336 and 335.

Regarding https://circleci.com/gh/status-im/react-native-desktop/566 it sounds like when you do `exit 1` script executor might exits explicitly as well. But normally it probably returns last command exit code (haven't verified that)

madhavarshney: > Indeed, cmake does according to following logs:
https://circleci.com/gh/MaxRis/react-native-desktop/335
https://circleci.com/gh/MaxRis/react-native-desktop/336
Please compare the output for cmake commands between 336 and 335.

My bad, that seems to be true.

> But normally it probably returns last command exit code (haven't verified that)

Not according to [the CircleCI docs](https://circleci.com/docs/2.0/configuration-reference/#default-shell-options).  Maybe I misunderstood the docs.

madhavarshney: Just found [this forum thread](https://discuss.circleci.com/t/tests-fail-but-build-continues-and-finishes-successfully/10450).  Seems like you're right, although the official docs somewhat contradict that thread.

madhavarshney: I think `set -e` in each step would probably be neater.

.circleci/config.yml
madhavarshney: @MaxRis - Just [hit this issue](https://circleci.com/gh/status-im/react-native-desktop/581?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) in #279.  `--force-yes` doesn't seem to necessarily solve this issue because in that PR `apt-add-repository` fails with a non-zero exit code.  This seems to be a [widespread problem/known issue](https://github.com/docker-library/mysql/issues/263#issuecomment-354025886) with GPG keyservers.  As I mentioned above, a prebuilt docker image would be best.

PR #282, opened 2018-07-03
update dimensions for desktop

rcullito:
this is meant to address: https://github.com/status-im/status-react/issues/4824

@MaxRis @vkjr are there any commands I should be running for these parameters to effect?

I reinstalled the local `react-native-cli` via npm, bu it seems like more steps may need to be taken. thanks!

### DIFF
--- a/ReactQt/application/src/main.qml.in
+++ b/ReactQt/application/src/main.qml.in
@@ -13,8 +13,8 @@ import React 0.1 as React
 
 Rectangle {
   id: root
-  width: 384
-  height: 640
+  width: 600
+  height: 500
 
   React.RootView {
     anchors.fill: parent


madhavarshney:
> are there any commands I should be running for these parameters to effect?
I reinstalled the local react-native-cli via npm, bu it seems like more steps may need to be taken. thanks!

Check https://github.com/status-im/react-native-desktop/pull/276#issuecomment-400747106 and https://github.com/status-im/react-native-desktop/pull/276#issuecomment-400757068.  By the way, CI is failing because this PR is not based on the latest master.

MaxRis:
@rcullito as Madhav mentioned you need to update index.js of your `react-native-cli` package to point updated remote git branch with your changes for the next `react-native init` command you will do.
Or as much more easy alternative you can just modify your `package.json` to point to the desired react-native branch, optionally clean node_modules directory, and install all modules or only react-native.

MaxRis:
@rcullito  Regarding the changes itself, it seems that some API should be provided to change start size of the main window (but I have a feeling that we already can do this from `desktop/main.cpp` directly in C++ code)

vkjr:
@rcullito, in general, your changes will work. But probably this is not the best approach. By changing dimensions in the react-native-desktop itself we force all apps based on react-native-desktop to have the same dimensions.
Instead, users of react-native-desktop should be able to have own `main.qml`  file and modify it as needed. Similarly, like it is done right now with `main.cpp`: all apps based on react-native-desktop have a default one, but you also can have a modified one that is saved in `desktop` folder of your project.

madhavarshney:
I agree with both @MaxRis and @vkjr.  In my opinion, two useful things to do would be to document that this is possible and _maybe_ increase the initial size of the app to a more desktop-friendly size (which would mean updating the samples as well).

rcullito:
thanks for the feedback here guys. closing this for now in favor of resizing within `status-react` repo

Issue #291, opened 2018-07-10
Use global react-native-cli directly


madhavarshney:
@MaxRis @vkjr - Today I discovered that you don't need to use the custom cli of `react-native-desktop` to create a new project and you can directly use the stock `react-native-cli` and specify the version of react-native to use instead.  The commands to create a new project would be:

```
npm install --global react-native-cli
react-native init RNDesktopTest --version status-im/react-native-desktop
cd RNDesktopTest
react-native desktop
```

I can submit a PR to add this method to the docs and remove the other one, as this is somewhat easier, removing the need of cloning the whole `react-native-desktop` repository.

MaxRis:
@madhavarshney nice finding! Docs update sounds great!

PR #292, opened 2018-07-10
Recommend using stock react-native-cli in docs instead of modified version

madhavarshney:
As described in #291.

### DIFF
--- a/docs/InstallUpdatedReactNativeCLI.md
+++ b/docs/InstallUpdatedReactNativeCLI.md
@@ -1,9 +1,9 @@
 ## Overview
 
-Original `react-native-cli` is extended with cli commands to generate, build and run react-native projects for `Desktop` platform. To be able to create react-native apps for `Desktop` platform you should install extended version of cli by following next steps:
+Original `react-native-cli` is extended with CLI commands to generate, build and run react-native projects for `desktop` platform. To be able to create react-native apps for `desktop` platform you should either specify `--version` in the original CLI when creating a react-native app, `react-native init DesktopApp --version status-im/react-native-desktop`, or install the extended version of the CLI by following the next steps and then running `react-native init DesktopApp`.
 
-1. git clone https://github.com/status-im/react-native-desktop.git
-2. cd react-native-desktop/react-native-cli
-3. npm update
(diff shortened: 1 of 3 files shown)


docs/InstallUpdatedReactNativeCLI.md
```diff
@@ -1,9 +0,0 @@
-## Overview
```
vkjr: @madhavarshney, I tried the approach you mentioned. Works great, thanks!
I suggest not to remove this document, just leave a link to in in `CreateNewApp.md`, as to alternative way.

madhavarshney: I don't think it's needed because the other approach is much easier and has greater control over the version of `react-native-desktop` to install.  Would you mind elaborating on when this can be useful?  The one case I can think of is when you don't want to specify `--version`, but that's more of an edge case in my opinion.

vkjr: @madhavarshney, I think about a bright beautiful future when some people develop a lot of desktop apps on the same machine and want to have adjusted `react-native-cli` installed to type fewer characters on project initialization :)
Also, I think that if adjusted `react-native-cli` can be installed globally, we should at least mention this possibility to not forget that it exists :)

madhavarshney: In one sense, yes.  However, when we refactor the repository to be an extension of `react-native` rather than a fork and full implementation, the adjusted global CLI wouldn't exist, and instead `rnpm-plugin-desktop` would have to be installed in whichever repository `desktop` support needs to be added to.  For now I will leave the file as it is.

vkjr: @madhavarshney, agree, when we refactor the repository we can get rid of this part of documentation since it won't be usable at all. But for now please leave this doc and link to it, "just in case".

madhavarshney: Done, please take a look.

MaxRis:
@madhavarshney thank you! Awesome finding!

vkjr:
@madhavarshney, thank you for your input!

PR #294, opened 2018-07-14
react-native-sound-level added

punarinta:


punarinta:
@MaxRis, the only issue was a Qt-related one, I mentioned it in the intallation manual: Qt installer hadn't linked one of the libraries and so linker didn't see it while building the project

vkjr:
@punarinta, do you plan to use `react-native-desktop` to create desktop apps? We'd be glad to hear about successful cases :) Or about issues that need to be fixed :)

punarinta:
@vkjr, not yet, but I'm experimenting. I'd say sending events from a native layer to js would be quite useful to have before building complex apps. If that's already possible I would be very thankful if you could point out how to do that. For example I'd like to send geolocation data when it's updated. Asking for it explicitly on js timer updates looks like an ugly hack to me.

punarinta:
Got it. Thank you! I was indeed confused by that naming.

vkjr:
@punarinta, you are welcome. Please create an issue if that approach won't work :)

MaxRis:
@punarinta @vkjr there is seems exists more "standard" way to receive native notifications in JS. As example of it, please, check implementation of `./Libraries/Network/NetInfo.js` module. To receive native events from `NetInfo` native module following construction is done there:

```
const NativeEventEmitter = require('NativeEventEmitter');
const NativeModules = require('NativeModules');
const Platform = require('Platform');
const RCTNetInfo = NativeModules.NetInfo;

const NetInfoEventEmitter = new NativeEventEmitter(RCTNetInfo);

const DEVICE_CONNECTIVITY_EVENT = 'networkStatusDidChange';
```

and event listener subscription is done

```
listener = NetInfoEventEmitter.addListener(
        DEVICE_CONNECTIVITY_EVENT,
        (appStateData) => {
          handler({
            type: appStateData.connectionType,
            effectiveType: appStateData.effectiveConnectionType
          });
        }
      );
```

netinfo.cpp native module implementation emits required event as following:

```
bridge->eventDispatcher()->sendDeviceEvent(
                                 "networkStatusDidChange",
                                 QVariantList{QVariantMap{{"network_info", accessibleName.value(accessible)}}});
                         }
```

Hopefully, similar approach can be used to emit/receive any custom event for any native module.

vkjr:
@MaxRis, thanks!

PR #305, opened 2018-07-29
Switch component implementation

### LINKED ISSUE #293: "Switch" component support
["Switch"](https://facebook.github.io/react-native/docs/switch) react-native component should be supported on desktop.

vkjr:
Fixes #293 
Added `Switch` component. 

Screenshot from RNTester:
screen shot 2018-07-29 at 5 07 16 pm

### DIFF
--- a/ReactQt/runtime/src/bridge.cpp
+++ b/ReactQt/runtime/src/bridge.cpp
@@ -583,7 +583,7 @@ void Bridge::invokeModuleMethod(int moduleId, int methodId, QList<QVariant> args
     }
 
     // readable log of methods invoked via bridge
-    //    qDebug() << "INVOKE: " << moduleData->name() << "::" << method->name() << "( " << args << " )";
+    // qDebug() << "INVOKE: " << moduleData->name() << "::" << method->name() << "( " << args << " )";
 
     method->invoke(args);
 }
--- a/ReactQt/runtime/src/bridge.cpp
+++ b/ReactQt/runtime/src/bridge.cpp
@@ -31,6 +31,7 @@
 #include "componentmanagers/rawtextmanager.h"
 #include "componentmanagers/scrollviewmanager.h"
 #include "componentmanagers/slidermanager.h"
+#include "componentmanagers/switchmanager.h"
 #include "componentmanagers/textinputmanager.h"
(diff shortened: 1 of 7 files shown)


ReactQt/runtime/src/qml/ReactSwitch.qml
```diff
@@ -0,0 +1,79 @@
+import QtQuick 2.9
+import QtQuick.Controls 2.2
+import React 0.1 as React
+import QtQuick.Controls.impl 2.2
```
MaxRis: pretty cool! Seems we should use similar approach to land others controls with default styling, but it seems I was not able to find any documentation of QtQuick.Controls.impl

ReactQt/runtime/src/bridge.cpp
```diff
@@ -581,7 +583,7 @@ void Bridge::invokeModuleMethod(int moduleId, int methodId, QList<QVariant> args
     }
 
     // readable log of methods invoked via bridge
-    // qDebug() << "INVOKE: " << moduleData->name() << "::" << method->name() << "( " << args << " )";
+    //    qDebug() << "INVOKE: " << moduleData->name() << "::" << method->name() << "( " << args << " )";
```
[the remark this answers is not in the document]
vkjr: Do you mean whitespaces or the commented code itself?

I'd prefer to leave that commented line in the code because it is extremely helpful. I'm uncommenting it almost every time during development :)

ReactQt/runtime/src/qml/ReactSwitch.qml
vkjr: I haven't seen any documentation also. `QtQuick.Controls.impl` here needed only for default colors constants.
I took that from Switch.qml and SwitchComponent.qml original implementation.

ReactQt/runtime/src/bridge.cpp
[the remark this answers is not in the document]
MaxRis: Yes, only whitespaces :)

ReactQt/runtime/src/qml/ReactSwitch.qml
MaxRis: Maybe Slider control customization can use something like this

Issue #303, opened 2018-07-28
Move network communications to separate thread


MaxRis:
To avoid overloading of main thread with network communications it makes sense to move Executor to separate thread

PR #306, opened 2018-07-29
Move Executor to separate thread

MaxRis:
Closes https://github.com/status-im/react-native-desktop/issues/303

Executor and Connection instances are moved to separate QThread.

`invokeModuleMethod` call happens back in main thread, since it calls native modules created on main thread and should be executed on main thread because of this.

Inter-thread communication happens via QMetaObject::invokeMethod with Qt::AutoConnection to post the call into the event loop of invoked object's thread automatically.

Testing notes for status-react Desktop:

Application should operate as usual: Login, Contact adding, send a message, etc.

### DIFF
--- a/ReactQt/runtime/src/bridge.cpp
+++ b/ReactQt/runtime/src/bridge.cpp
@@ -66,6 +66,7 @@
 #include <QPluginLoader>
 #include <QQuickItem>
 #include <QStandardPaths>
+#include <QThread>
 #include <QTimer>
 
 class BridgePrivate {
@@ -89,6 +90,7 @@ public:
     bool remoteJSDebugging = false;
     bool hotReload = false;
     QVariantList externalModules;
+    QThread* executorThread = nullptr;
 
     bool useJSC = false;
 
@@ -128,7 +130,9 @@ Bridge::Bridge(QObject* parent) : QObject(parent), d_ptr(new BridgePrivate) {
     d->eventDispatcher = new EventDispatcher(this);
 }
 
-Bridge::~Bridge() {}
+Bridge::~Bridge() {
+    resetExecutor();
+}
 
 void* Bridge::getJavaScriptContext() {
     Q_D(Bridge);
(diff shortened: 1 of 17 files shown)


vkjr:
@MaxRis, did you tried this version of `react-native-desktop` on `status-react` with heavy chat loading? ?

MaxRis:
@vkjr yes, I've tried to test it, but have not seen some stable or visible performance improvement neither for chat loading nor for a lot messages rendering on scrolling.

Probably, we should also aware that QML scene rendering happens in separate thread and blocking main thread only partially ( http://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph.html#threaded-render-loop-threaded ). UI responsiveness lost is just a result of too much ui elements on the scene graph.

All to all moving Executor to separate thread should have positive impact for apps with much higher data transfer volume between JS and native part than we have in status-react currently. ( I believe, that Objective-C has RCTMessageThread implementation which keeps Executor as well ).

-----------------

Recently have found that react-native has built-in optimization for Views to not render ui elements currently not visible. Looking into how it works and how to add similar implementation of this into our native code.

```
/**
     * Experimental: When true, offscreen child views (whose `overflow` value is
     * `hidden`) are removed from their native backing superview when offscreen.
     * This can improve scrolling performance on long lists. The default value is
     * true.
     */
    removeClippedSubviews: PropTypes.bool,
```

MaxRis:
Jenkins build with react-native-desktop changes from this PR: https://jenkins.status.im/blue/organizations/jenkins/status-react%2Fdesktop%2Fmanual/detail/manual/56/artifacts/

vkjr:
Thanks!
Great finding!

MaxRis:
Just note: Have found that the simplest way to measure manually UI responsiveness with opened chat history is to keep mouse button pressed on some message text and actively change text selection. On chats with low amount of messages in history selection works smoothly, but if more messages are loaded from the history responsiveness starts to degrade more and more.

PR #310, opened 2018-08-07
ScrollView content rendering optimization

MaxRis:
**Issue description:**

ScrollView keeps all children views visible in scene graph and ui experiences responsiveness issues on huge amount of visual items. 

**Changes description:**

Changes in PR introduces new props into ScrollView.js component:

```
     /**
     * Enables rendering optimization for content's items arrays inside scroll view
     * @platform desktop
     */
     enableArrayScrollingOptimization: PropTypes.bool,
    /**
     * Scroll view content top padding.
     * `enableArrayScrollingOptimization` prop should be enabled.
     * @platform desktop
     */
     headerHeight: PropTypes.number,
    /**
     * Scroll view content bottom padding.
     * `enableArrayScrollingOptimization` prop should be enabled.
     * @platform desktop
     */
     footerWidth: PropTypes.number,
```

When `enableArrayScrollingOptimization` is set to `true`, ScrollViewManager expects direct child of ScrollView to be defined as array of elements. Newly introduced ReactScrollListView.qmll is ListView qml component based. ListView takes the model of pre-created QQuickItem views of array's elements and mount/unmount them dynamically into scene graph (this happens in ListView's delegate component). Additionally, with `headerHeight` and `footerWidth` props its possible to set top and bottom padding for elements in ListView.

### DIFF
--- a/ReactQt/runtime/src/componentmanagers/scrollviewmanager.cpp
+++ b/ReactQt/runtime/src/componentmanagers/scrollviewmanager.cpp
@@ -15,6 +15,7 @@
 #include <QQuickItem>
 #include <QString>
 #include <QVariant>
+#include <QVariantList>
 
 #include <QDebug>
 
@@ -29,6 +30,9 @@
 
 using namespace utilities;
 
+QMap<QQuickItem*, QQuickItem*> ScrollViewManager::m_scrollViewByListViewItem;
+QMap<QQuickItem*, QVariantList> ScrollViewManager::m_modelByScrollView;
+
 void ScrollViewManager::scrollTo(int reactTag, double offsetX, double offsetY, bool animated) {
     QQuickItem* item = bridge()->uiManager()->viewForTag(reactTag);
     Q_ASSERT(item != nullptr);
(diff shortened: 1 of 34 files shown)


vkjr:
@MaxRis, great PR, thank you!
Have you checked it already with `status-react`? Any noticeable changes?

ReactQt/runtime/src/componentmanagers/scrollviewmanager.cpp
```diff
@@ -153,8 +219,14 @@ void ScrollViewManager::configureView(QQuickItem* view) const {
     connect(view, SIGNAL(flickEnded()), SLOT(momentumScrollEnd()));
 }
 
-QString ScrollViewManager::qmlComponentFile() const {
-    return "qrc:/qml/ReactScrollView.qml";
+QString ScrollViewManager::qmlComponentFile(const QVariantMap& properties) const {
+
+    return properties.value("enableArrayScrollingOptimization", false).toBool() ? "qrc:/qml/ReactScrollListView.qml"
```
vkjr: @MaxRis, do we need to support both type of ScrollViews? Maybe we can go with new one in all cases?

vkjr:
There is a property `removeClippedSubviews` in original ScrollView. Does it results in any additional calls from js via bridge? I was hoping that js already takes care about optimization and notifies native part when some components should be unloaded. Or it is not the case?

ReactQt/runtime/src/componentmanagers/scrollviewmanager.cpp
MaxRis: @vkjr that's a drawback of current optimization. It will work if array of children has ScrollView as a direct parent:
<ScrollView> <ArrayOfScrollableChildren /> </ScrollView>

and will not work in any other case, for example:
<ScrollView> <View><ArrayOfScrollableChildren /></View> </ScrollView>

MaxRis:
@vkjr performance for chats with long messages history should be increased dramatically. You can check test build there https://ci.status.im/blue/organizations/jenkins/status-react%2Fdesktop%2Fmanual/detail/manual/121/artifacts
Also there seems was issue with top left ScrollView content point calculation ( it should depend on originalY property as well, not only on contentY ). Loading of the previously received messages from the database seems works much better also (when you scrolled to the top of history).

PR #311, opened 2018-08-08
wrap mode changed for text

vkjr:
Change wrapping mode in Text, so it wrap text by words, not in an arbitrary place.

### DIFF
--- a/ReactQt/runtime/src/qml/ReactText.qml
+++ b/ReactQt/runtime/src/qml/ReactText.qml
@@ -41,7 +41,7 @@ TextEdit {
     text: textIsTopInBlock ? decoratedText : ""
     horizontalAlignment: horizontalAlignmentFromTextAlign(p_textAlign)
     textFormat: Text.RichText
-    wrapMode: Text.WordWrap
+    wrapMode: Text.Wrap
     clip: true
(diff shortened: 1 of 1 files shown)


MaxRis:
@vkjr just curious, if `QTextOption::WrapAtWordBoundaryOrAnywhere` is also available from there. It might be more appropriate.
Another thing, if react-native has special props to handle text wrapping mode, we need to support it from JS level

vkjr:
@maxris,
```
TextEdit.NoWrap - no wrapping will be performed. If the text contains insufficient newlines, then implicitWidth will exceed a set width.
TextEdit.WordWrap - wrapping is done on word boundaries only. If a word is too long, implicitWidth will exceed a set width.
TextEdit.WrapAnywhere - wrapping is done at any point on a line, even if it occurs in the middle of a word.
TextEdit.Wrap - if possible, wrapping occurs at a word boundary; otherwise it will occur at the appropriate point on the line, even in the middle of a word.
```

Good point, probably `TextEdit.Wrap` is similar to `QTextOption::WrapAtWordBoundaryOrAnywhere`, will use it.

As for specific property, I checked Text component and didn't find it.

PR #312, opened 2018-08-08
Add methods to check os name from js

vkjr:
`Platform` module extended, now you can check what os you are using.
That will be helpful when we select for fonts to use in `status-react` depending on Linux or Mac.

Example of usage:
```
changeButtonText = () => {
        if(Platform.isMacOs){
            this.setState({buttonText: 'macos'})
        }
        else{
            this.setState({buttonText: 'unknown'})
        }
      }
```

### DIFF
--- a/ReactQt/runtime/src/platform.cpp
+++ b/ReactQt/runtime/src/platform.cpp
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2017-present, Status Research and Development GmbH.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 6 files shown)


ReactQt/runtime/src/platform.cpp
```diff
@@ -0,0 +1,61 @@
...
+        osName = QLatin1String("unix");
+#else
+        osName = QLatin1String("unknown");
+#endif
+    }
+
+public:
+    Bridge* bridge = nullptr;
+    QString osName;
+    int major = 0;
```
vkjr: Original `react-native` uses `PlatformConstants` module to store not only os type constants, but also react-native version for runtime checks, so I added it here also.

PR #324, opened 2018-08-31
Fix svg blurriness

### LINKED ISSUE #322: SVG images are blurry
**Description**

The following code exhibits the problem:
```
import React, {Component} from 'react';
import { AppRegistry, Text, View, TouchableHighlight, Image} from 'react-native';

class BasicExample extends Component {
  render () {
    return (
      <Image source={require('./warning.svg')}
             style={{width: 100, height: 100}}/>
);
  }
}

AppRegistry.registerComponent('rn_test', () => BasicExample);
```

`warning.svg` is taken from status-react repo ([link](
https://github.com/status-im/status-react/blob/develop/resources/icons/warning.svg)).

vkjr:
Fixes #322 

Svg images sometimes contain internal height and width specification and get blurry when resized. To avoid this we can set sourceSize that supercedes internal svg settings.

### DIFF
--- a/ReactQt/runtime/src/componentmanagers/imagemanager.cpp
+++ b/ReactQt/runtime/src/componentmanagers/imagemanager.cpp
@@ -77,6 +77,8 @@ void ImageManager::manageSource(const QVariantMap& imageSource, QQuickItem* imag
         return;
     }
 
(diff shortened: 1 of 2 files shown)


vkjr:
@siphiuel, thanks for checking with status-react!

PR #327, opened 2018-09-04
Log yoga nodes managing

MaxRis:
In scope of efforts to catch yoga nodes related exceptions PR introduces logging of yoga nodes adding, removing and details of uimanager::manageChildren invocation.


MaxRis:
@vkjr yes, good idea!

PR #330, opened 2018-09-07
Support inverted property for FlatList component

vkjr:
`FlatList` component declared by react-native as an optimized scroll view that partially unloads items when they are out of visible area. In is based on `ScrollView`, so `react-native-desktop` works fine with it. 
The idea was to use FlatList in `status-react` desktop the same way mobile team does. To make scrolling without jumps in FlatList, mobile team uses `inverted` property. 

This PR goal is to fix `inverted` property to work correctly in desktop `FlatList`. Initially it didn't work because for inverting `Flatlist` code uses transformation that wasn't implemented in desktop `ScrollView`. Fixing it also revealed issue: when vertical transformation applied, Qt wheel events works  incorrectly and change scroll direction to opposite, so this was additionally fixed.

### DIFF
--- a/ReactQt/runtime/src/qml/ReactScrollView.qml
+++ b/ReactQt/runtime/src/qml/ReactScrollView.qml
@@ -11,6 +11,16 @@ Flickable {
     property bool p_enableArrayScrollingOptimization: false
     property int p_headerHeight: 0
     property int p_footerWidth: 0
+    property var p_transform
+    property bool invertedScroll: false
+
+    onP_transformChanged: {
+        scrollViewManager.addTransformation(scrollViewRoot, p_transform)
+        //if scrollview vertically inveted we should remember that to adjust wheel events
+        var scaleY = p_transform[5];
+        invertedScroll = (scaleY < 0);
+    }
+
 
     clip: true
     contentHeight: contentItem.childrenRect.height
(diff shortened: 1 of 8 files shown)


ReactQt/runtime/src/qml/ReactScrollView.qml
```diff
@@ -11,6 +11,16 @@ Flickable {
     property bool p_enableArrayScrollingOptimization: false
     property int p_headerHeight: 0
     property int p_footerWidth: 0
+    property var p_transform
+    property bool invertedScroll: false
+
+    onP_transformChanged: {
+        scrollViewManager.addTransformation(scrollViewRoot, p_transform)
+        //if scrollview vertically inveted we should remember that to adjust wheel events
+        var scaleY = p_transform[5];
```
MaxRis: maybe p_transform should be validated somehow before usage here

MaxRis:
From [react-native's doc](https://facebook.github.io/react-native/docs/flatlist):
```
In order to constrain memory and enable smooth scrolling, content is rendered asynchronously offscreen. This means it's possible to scroll faster than the fill rate and momentarily see blank content. This is a tradeoff that can be adjusted to suit the needs of each application, and we are working on improving it behind the scenes.
```
We might should aware about what happens `behind the scenes`, because it might be the only way to optimize rendering for Desktop on JS-only side. ( There might be some kind of rendering offset property to play with)

PR #333, opened 2018-09-12
Model added to scrollview

vkjr:
- Model class implemented for ScrollView
- inverted property added to ScrollView

Having inverted property implemented allows us to avoid chat jumping in status-react when new messages loaded from history

### DIFF
--- a/Libraries/Components/ScrollView/ScrollView.js
+++ b/Libraries/Components/ScrollView/ScrollView.js
@@ -494,6 +494,11 @@ const ScrollView = createReactClass({
        PropTypes.number,
      ]),
     /**
+     * Inverts layout of elements within scrollview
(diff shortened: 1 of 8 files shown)


Libraries/Components/ScrollView/ScrollView.js
```diff
@@ -920,6 +925,7 @@ const ScrollView = createReactClass({
       }
     }
     if (Platform.OS === 'desktop' && props.enableArrayScrollingOptimization) {
+      console.log("props: ", {...props})
```
MaxRis: Probably, should be removed

vkjr: For sure! Thanks for noticing :)

PR #354, opened 2018-10-02
Selective logging added

### LINKED ISSUE #329: Selective logging in react-native-desktop
Currently `react-native-desktop` produces quite a lot of log output to console.
I'd be happy if we have some configuration file where logging can be tweaked, something like:

```
LOG_FLEX=TRUE
LOG_NETWORK=FALSE
LOG_STATUSGO=FALSE
LOG_BRIDGE_EVENTS=TRUE
```

That would make reading logs more convenient

vkjr:
Fixes #329

- Added logger that can be turned on/off for specific module. 
- Replaced qDebug() calls with new logger in Flexbox, ViewManager, UIManager, Networking, WebSocketModule.

Logger creates `logger_settings.ini` file in a `desktop` directory of `react-native-desktop` project. By changing settings in that file you can turn on and off logging for different modules. To pickup changes in `logger_settings.ini` reloading required (but not restarting).

This version of logger works "as is" with 3rd party modules. PR added to add RCTStatus support - https://github.com/status-im/status-react/pull/6086

### DIFF
--- a/ReactQt/runtime/src/logger.cpp
+++ b/ReactQt/runtime/src/logger.cpp
@@ -1,0 +1,74 @@
+
+/**
+ * Copyright (c) 2017-present, Status Research and Development GmbH.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include "logger.h"
+#include <QBuffer>
+
+const QString LOG_SETTINGS_FILE = "logger_settings.ini";
+
+Logger& Logger::getInstance() {
+    static Logger logger;
+    return logger;
+}
+
+Logger::~Logger() {
+    delete m_settings;
+    m_settings = nullptr;
+}
+
+QString& Logger::devNull() {
+    return m_devnull;
+}
+
+const QSettings& Logger::settings() {
+    return *m_settings;
+}
+
+void Logger::registerModule(const QString& moduleName) {
+    if (!m_settings->contains(moduleName)) {
+        m_settings->setValue(moduleName, true);
+    }
+}
+
+void Logger::sync() {
+    m_settings->sync();
+}
+
+Logger::Logger() {
+    m_settings = new QSettings(LOG_SETTINGS_FILE, QSettings::IniFormat);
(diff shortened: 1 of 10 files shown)


ReactQt/runtime/src/logger.cpp
```diff
@@ -0,0 +1,74 @@
...
+    registerModule(UIMANAGER);
+    registerModule(FLEXBOX);
+    registerModule(VIEWMANAGER);
+}
+
+QDebug rnLog(const QString& module) {
+    Logger& logger = Logger::getInstance();
+    logger.devNull().clear();
+
+    if (!logger.settings().contains(module)) {
```
MaxRis: curios, if `rnLog` is thread-safe method. At the moment we have: main thread, thread with JS engine communications, status-go thread in status-react rctstatus module, qtconcurrent calls from rctstatus module.

ReactQt/runtime/src/logger.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include "logger.h"
+#include <QBuffer>
+
+const QString LOG_SETTINGS_FILE = "logger_settings.ini";
```
MaxRis: As we discussed, introduction of another one settings file (similar to .env) should create additional complexity to manage settings for status-react and others derived projects.
Probably, for future, what if to embed support of react-native-config and .env files directly into react-native-desktop on native level.

ReactQt/runtime/src/logger.cpp
vkjr: I believe it should be ok:
- Mayers singleton is thread-safe in C++11 (https://stackoverflow.com/questions/1661529/is-meyers-implementation-of-the-singleton-pattern-thread-safe)
- QSettings methods are reentrant
- clearing of `m_devnull` has no sideeffects because data inside that string never used
- qDebug() also seems to be safe

MaxRis: @vkjr thank you for checking!
P.S. When we install our custom handler for qDebug(), we should care about thread safety of handler implementation (at least that was the case with status-react).

PR #361, opened 2018-10-05
fix defaultValue prop in TextInput

rasom:
Attempt to fix behavior of `defaultValue` property of `TextInput`.

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -812,6 +812,7 @@ const TextInput = createReactClass({
   _inputRef: (undefined: any),
   _focusSubscription: (undefined: ?Function),
   _lastNativeText: (undefined: ?string),
+  _initialDefaultValue: (undefined: ?string),
   _lastNativeSelection: (undefined: ?Selection),
 
   componentDidMount: function() {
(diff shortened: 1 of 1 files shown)


vkjr:
@rasom, to be honest, I don't think that fixing react-native itself is an ideal solution. Could you please describe what exactly goes wrong without that fix? Is the problem happens due to some specifics of `TextInput` use in clojurescript? If so, maybe we can fix on a clojurescript side. 
If the problem has also roots in `react-native-desktop` C++\Qt code, maybe we can fix there?

Do we have the same problem in mobile? If yes - do we make similar changes in `react-native` mobile uses? If no - what is the difference with desktop

Sorry for making troubles, just want to make sure we do right things :)

rasom:
> Sorry for making troubles, just want to make sure we do right things :)

oh, no @vkjr, that's exactly what I expected after filing this silly PR :D 

So, this PR makes `defaultValue` prop to behave exactly the way it works on mobile side atm (i mean from API client's pov, not exact same mechanics under the hood). We actually already tried to make a fix on cljs side, this doesn't make sense because we will have a discrepancy between the way `TextInput` is used on mobile and desktop. 

The problem is that `defaultValue` is applied each time when the component is updated, but it should only set an initial value for the `TextInput`. And when it is applied, for some reason caret position gets reset to 0. This causes https://github.com/status-im/status-react/issues/6074 .

I will be happy if fix change will be introduced on `C++\Qt`, but depending on the amount of time necessary for this change we could use this hack meanwile because [this bug](https://github.com/status-im/status-react/issues/6074) is quite annoying and easily reproducible on slower computers.

vkjr:
I'll check the Qt side this weekend. If no quick solution - will merge this.

rasom:
thank you @vkjr

vkjr:
@rasom, this PR makes `defaultValue` to work the way user suppose it works. So I think it would be good to merge it to master. But I think we shouldn't change mobile code because someone may want to base already existing project on `react-native-desktop`.
I'd suggest to create new function `getTextDesktop` (or any name) with your changes and use it inside `_renderDesktop`. And leave original `_getText` untouched, to not affect ios/android code. wdyt?

rasom:
ok, will do this

rasom:
@vkjr done

vkjr (approved):

Issue #214, opened 2018-04-20
TextInput events repeats constantly


vkjr:
Steps to reproduce:
- open RNTester app
- in search textinput quickly type few symbols
Result:
- symbols will appear and disappear constantly.

vkjr:
Fixed.

PR #365, opened 2018-10-08
Fixed double sending of onTextChange events

vkjr:
Two TextInput bugs fixed:
1) now TextInput doesn't send "onChange" events when text was set from js side. That solves bug #214
2) I found a discrepancy between multiline and singleline TextInput in Qt. When new text assigned, singleline TextInput saves cursor position, whereas multiline one - doesn't. In second case that results in cursor jumping when we quickly type in text because typed text sent to js and then assigned again from js to Qt because `defaultValue` changed.

PR to check in status-react: https://github.com/status-im/status-react/pull/6176
Here is a code that reveals second problem in js:
```
export default class ButtonReactNative extends Component {

  constructor(props) {
    super(props);
    this.state = { newtext: "" };
  }

  render() {
    return (
      <TextInput
        style={{ backgroundColor: "yellow", width: 300, height: 100 }}
        defaultValue={this.state.newtext}
        multiline={true}
        onChangeText={text => {
          this.setState({ newtext: text })
          console.log("new text: ", text)}
        }
      />
    );
  }
}
```

### DIFF
--- a/Libraries/Components/TextInput/TextInput.js
+++ b/Libraries/Components/TextInput/TextInput.js
@@ -172,10 +172,15 @@ type AndroidProps = $ReadOnly<{|
   inlineImagePadding?: ?number,
 |}>;
 
+type DesktopProps = $ReadOnly<{|
+  submitShortcut?: ?object,
+|}>;
+
 type Props = $ReadOnly<{|
   ...ViewProps,
   ...IOSProps,
   ...AndroidProps,
+  ...DesktopProps,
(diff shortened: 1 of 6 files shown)


maxhora (approved):
Nice catch!

yenda:
Awesome ! What's wrong with the circle ci ?

vkjr:
@yenda, we have one of standard `react-native` tests failing (because not implemented). Max recently temporary excluded it from test suite

rasom (approved):

PR #368, opened 2018-10-09
Don't remove scroll view model item on get request

maxhora:


vkjr (approved):

vkjr:
What kind of bug happens due to this removal before fix?

MaxRis:
@vkjr I was getting crashes when tried to apply optimization for chats list https://github.com/status-im/status-react/pull/6229
It happened on move from/to operation for optimized scroll-view. Well, it really makes sense to cover optimized scroll-view with similar set of unit-tests.

PR #369, opened 2018-10-11
Fix/memory leaks

### LINKED ISSUE #348: Memory leaks on re-rendring FlatList
It seems that when FlatList is unmounted memory is not released, which causes memory leaks if you have to render/unmount lists with many elements frequently.

Sample code:
```js
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Text,
  TextInput,
  View, 
  FlatList
} from 'react-native';

var coll1 = [];
var coll2 = [];
for(i = 0; i < 100; i++) {
    coll1[i] = {key: 'a' + i};
    coll2[i] = {key: 'b' + i};
}

function List(props) {
    const s = props.switch;
    if (s) {
        return <FlatList
            style={{top:40, height: 1000}}
            data={coll1}
            initialNumToRender={100}
            renderItem={({item}) => <Text>{item.key}</Text>}
        />
    } else {
        return <FlatList
            style={{top:40, height: 1000}}
            data={coll2}
            initialNumToRender={100}
            renderItem={({item}) => <Text>{item.key}</Text>}
        />
    };
}


class App extends React.Component {
    constructor(props) {
        super(props);
        this.state = { switch: true, };
    }

    render() {
        return (
                <View>
                    <Text onPress={() => this.setState({switch: this.state.switch ? false: true})}>
                        click me to render another list
                    </Text>
                </View>
                <List switch={this.state.switch}/>
            </View>
        );
    }
}

const styles = StyleSheet.create({
    container: {
        flex: 1,
        backgroundColor: '#fff',
        alignItems: 'center',
        justifyContent: 'center',
    }
});

AppRegistry.registerComponent('DesktopSampleApp', () => App);
```

vkjr:
Partially fixes #348

Qml visual hierarchy differs from memory management hierarchy. So when we add new component it should be added to both. Then after deletion of top element, all childs also removed.

Steps to test with example in #348:
- run app (83Mb consumes at start)
- click on "click me" 30 times
result before: 240Mb consumed
result after: 97Mb consumed

### DIFF
--- a/ReactQt/runtime/src/uimanager.cpp
+++ b/ReactQt/runtime/src/uimanager.cpp
@@ -90,28 +90,31 @@ void UIManager::setChildren(int containerReactTag, const QList<int>& childrenTag
     manageChildren(containerReactTag, QList<int>(), QList<int>(), childrenTags, indices, QList<int>());
 }
 
-void UIManager::removeChildren(QQuickItem* parent, const QList<int>& removeAtIndices, bool unregisterAndDelete) {
+QList<QQuickItem*> UIManager::removeChildrenFromVisualParent(QQuickItem* parent, const QList<int>& removeAtIndices) {
 
     Q_ASSERT(parent != nullptr);
 
+    QList<QQuickItem*> itemsToRemove;
     if (!removeAtIndices.isEmpty()) {
 
-        QList<QQuickItem*> itemsToRemove;
         for (int i : removeAtIndices) {
             itemsToRemove.push_back(parent->childItems().at(i));
         }
 
         for (QQuickItem* child : itemsToRemove) {
             child->setParentItem(nullptr);
-
-            if (unregisterAndDelete) {
-                child->setParent(nullptr);
-                stopTrackingTagsForHierarchy(child);
-                child->deleteLater();
-            }
         }
 
         utilities::removeFlexboxChilds(parent, removeAtIndices);
+    }
+    return itemsToRemove;
+}
+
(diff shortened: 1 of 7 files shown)


ReactQt/runtime/src/uimanager.cpp
```diff
@@ -105,9 +105,8 @@ void UIManager::removeChildren(QQuickItem* parent, const QList<int>& removeAtInd
             child->setParentItem(nullptr);
 
             if (unregisterAndDelete) {
-                int childTag = AttachedProperties::get(child)->tag();
-                child->setParent(0);
-                m_views.remove(childTag);
+                child->setParent(nullptr);
```
MaxRis: @vkjr should this fix also be applied to ListView based ScrollView there https://github.com/status-im/react-native-desktop/blob/master/ReactQt/runtime/src/componentmanagers/scrollviewmanager.cpp#L109 ?

vitvly (approved):

ReactQt/runtime/src/uimanager.cpp
vkjr: Yes, good point!

yenda (approved):
Good job !

janherich (approved):

flexsurfer (approved):

yenda:
Merge and add to status-react quick please my RAM is crying :D

vkjr:
@maxris, I updated code to cover also ScrollView children removing.

maxhora (approved):
@vkjr thank you!

PR #370, opened 2018-10-16
Support nested onPress for Text components

### LINKED ISSUE #290: `onPress` handler doesn't work properly in nested `Text`
### Description
In `react-native` `Text` components can be nested and every nested element can have own `onPress` handler. In `react-native-desktop` onPress handler now works only for top-level element.

### Example
```
<Text onPress={()=>{console.log("Outer press")}}>
  Outer text
  <Text onPress={()=>{console.log("Inner press")}}>
    Inner text
  </Text>
</Text>
```

#### Expected behavior:
Click on `Outer text` should invoke `Outer press` console.log call.
Click on `Inner text` should invoke `Inner press` console.log call.

#### Actual behavior:
Both clicks invoke `Outer press` console.log call

siphiuel:
Fixes #290


MaxRis:
@siphiuel could you please fix failure of CircleCI? At the moment it seems fails because of missing clang-format formatting on modified C++ code

ReactQt/runtime/src/componentmanagers/textmanager.cpp
```diff
@@ -130,6 +132,36 @@ QVariant TextManager::nestedPropertyValue(QQuickItem* item, const QString& prope
...
 
+void TextManager::click(QQuickItem* textItem) {
+    qDebug() << "### click received";   
+
+    QPointF lp(0,0);
+    QPointF local(0,0);
+
+    QVariantMap e;
+    AttachedProperties* ap = AttachedProperties::get(textItem, false);
+    e.insert("target", ap->tag());
```
vkjr: Filling `QVariantMap` looks very similar in `RootView::makeReactTouchEvent`. It would be nice to move common code to single function, for example `utilities::createTouchArgs(tag, lp, local)`

siphiuel: Yes, I've just added a comment to the function that it's a simplified version of makeReactTouchEvent(). Will refactor.

siphiuel:
@MaxRis - oh, sorry, fixed.

ReactQt/runtime/src/componentmanagers/textmanager.cpp
vkjr: @siphiuel, please also remove `console.log` and `qDebug()` calls before merging. And changes in `react-native-cli/index.js` :)

MaxRis:
@siphiuel seems failed with checking `.h` file formatting now :)

siphiuel:
@maxris thanks, so lame on my end. Went through all formatting errors reported by clang-format. However, it complains about `./ReactQt/runtime/src/layout/flexbox.cpp` - not sure why

siphiuel:
@vkjr - thanks for suggestion, refactored into a separate fn in `utilities` ns.

vkjr:
@siphiuel, btw, if you don't have automatic formatting set in Qt Creator, you can use `fix-code-formatting.sh` script in `react-native-desktop` root folder to make sure everything formatted properly.

siphiuel:
@vkjr - thank you! I already used it, but due to (most likely) different clang-format versions CircleCI still cannot build the project. This script does not return any formatting errors on my machine, but CircleCI does complain.

siphiuel:
@vkjr @MaxRis turns out that CircleCI uses Ubuntu 16.04 image which has clang-format-3.8 in its repos. Downloaded it and used it to format the files. Now it passes the check.

react-native-cli/index.js
```diff
@@ -258,7 +258,7 @@ function createProject(name, options) {
 }
 
 function getInstallPackage(rnPackage) {
-  var packageToInstall =  'https://github.com/status-im/react-native-desktop#master';
+  var packageToInstall =  'https://github.com/siphiuel/react-native-desktop#fix/nested-onpress-for-text';
```
MaxRis:
```suggestion
  var packageToInstall =  'https://github.com/status-im/react-native-desktop#master';
```

siphiuel: Right, sorry, removed the cli change.

PR #372, opened 2018-10-19
Add vertical scroll bar to ScrollViews

vitvly:
Small fix for some kind of scrollbar to appear.

### DIFF
--- a/ReactQt/runtime/src/qml/ReactScrollListView.qml
+++ b/ReactQt/runtime/src/qml/ReactScrollListView.qml
@@ -12,6 +12,8 @@ ListView{
     property int p_headerHeight: 0
     property int p_footerWidth: 0
     property bool p_inverted: false
+    property bool p_showsHorizontalScrollIndicator: true
+    property bool p_showsVerticalScrollIndicator: true
 
     clip: true
     highlightFollowsCurrentItem: false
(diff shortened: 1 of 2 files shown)


maxhora (changes requested):
I believe that scrollbars should be optional, probably, controlled with prop in JS.
@vkjr what do you think?

ReactQt/runtime/src/qml/ReactScrollListView.qml
```diff
@@ -53,4 +53,8 @@ ListView{
             }
         }
     }
+
+    ScrollBar.vertical: ScrollBar {
+        width: 10
```
MaxRis: It should also has great sense to make width customizable

vkjr:
@maxris, @siphiuel, ScrollView has `showsVerticalScrollIndicator` `showsHorizontalScrollIndicator` properties that can be used.
By default these properties set to `true`, but scrollbars visible only for couple seconds while interaction in progress. When dragging stops, they disappear. I believe it would be nice to have same behavior. @siphiuel, does qt scrollbar behaves similarly? And I'm wondering if it has some default width value? I also would be glad not to specify a fixed width but use default instead.

siphiuel:
@vkjr - yes, by default it's the same behaviour. There is a `policy` property (https://doc.qt.io/qt-5.11/qml-qtquick-controls2-scrollbar.html#policy-prop) that can be modified to change scroll bar behavior.

siphiuel:
@vkjr - yes, there is a default width value, so it can work with just `ScrollBar.vertical: ScrollBar {}`.

siphiuel:
Added support `showsHorizontalScrollIndicator` and `showsVerticalScrollIndicator` properties, similarly to stock ReactNative. Scrollbar `width` is not available for customization in iOS/Android versions, so maybe we can skip on implementing it.

vkjr (approved):

maxhora (approved):

siphiuel:
@vkjr Fixed. Thank you for patience :)

PR #376, opened 2018-10-28
Make app a GUI app when building bundle and add app icon

PombeirP:
This PR passes `WIN32` parameter to `add_executable` on Windows so that a GUI app is generated instead of a console app. It also adds an app icon to Windows and Linux.

Part of https://github.com/status-im/status-react/issues/5807


ReactQt/application/src/resources/windows/resource.h
```diff
@@ -0,0 +1,19 @@
+//{{NO_DEPENDENCIES}}
```
MaxRis: @PombeirP I think it's fine to put there `// clang-format off` to suppress clang-format tool check for that file only. CircleCI build currently fails with complains about formatting issue.

PR #382, opened 2018-11-06
Replace custom logger class with Q_LOGGING_CATEGORY

pedropombeiro:
## Motivation

This PR replaces `rnLog` with the standard `QLoggingCategory` class to sidestep the crash seen when using `rnLog`. Fixes #381

By using `QLoggingCategory`, we are no longer limited to `qDebug`, and instead can use any logging level we need.

The logger_settings.ini gets replaced with the standard `QT_LOGGING_RULES` env var. See more in http://doc.qt.io/qt-5/qloggingcategory.html#details

Added documentation to https://github.com/status-im/status.im/blob/develop/source/build_status/desktop.md

PR #384, opened 2018-11-07
Add log documentation section and fix some lint warnings

PombeirP:
Part of #382

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,12 +1,13 @@
 ### How do Qt and Node communicate?
+
 Qt app contains one type of communication - it sends js code and receives JSON response to process.
 
 When started it sends full JS code of an application.
 
 When running it sends short pieces of JS code to invoke React Native functions and/or notify about events.
 
-
 ### How would you best describe the role of the `react-native-desktop` repo?
+
 Also: what code should live there? A build tool? Housing build related code?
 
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -52,7 +57,22 @@ Here is a JSON result that we receive from function above:
...
+The app leverages [QLoggingCategory](http://doc.qt.io/qt-5/qloggingcategory.html#configuring-categories) to allow filtering of categories. If the app sets the default categories using [QLoggingCategory::setFilterRules](http://doc.qt.io/qt-5/qloggingcategory.html#setFilterRules), then the categories can only be overriden using the `QT_LOGGING_CONF` and `QT_LOGGING_RULES` environment variables.
+
+Order of evaluation:
+
+1. `[QLibraryInfo::DataPath]/qtlogging.ini`
+1. `QtProject/qtlogging.ini`
+1. `setFilterRules()`
+1. `QT_LOGGING_CONF`
+1. `QT_LOGGING_RULES`
+
```
vkjr: @PombeirP, thanks for doc.
I think we can add an example related to react-native-desktop to make it more clear:
```
QT_LOGGING_RULES="*VIEWMANAGER=false;NETWORKING=true"
```

Btw, is there a way to somehow list all categories used within the `react-native-desktop`? That could be helpful for users. If no, probably we need to list them in documentation.

PombeirP: Agree with the example. I don't think there's any way to list the categories, as that is spread all around the code.

Issue #383, opened 2018-11-07
Reorganize react-native-desktop documentation


vkjr:
Currently `react-native-desktop` documentation is  bit scattered. Some common documentation listed as clojure-related, platform-independent information named as related to Ubuntu, etc..
Goal of this issue is to cleanup and reorganize documentation to make it more useful for new users and potential contributors.

PR #386, opened 2018-11-07
add example of setting the CMAKE_PREFIX_PATH within root CMakeLists.txt

rcullito:
## Motivation

related to improved documentation around: https://github.com/status-im/react-native-desktop/issues/383

using QtCreator on Linux without this path explicitly set was causing compilation errors when attempting to build the project. If it is better to set this var elsewhere (or not at all) please let me know! 

## CATEGORY
[ DOCS]

### DIFF
--- a/Development-linux.md
+++ b/Development-linux.md
@@ -30,6 +30,14 @@ Install:
 Launch `Qt Creator`, select `Tools`->`Options..`->`Build & Run`->`CMake` tab.
 Make sure that `CMake` tool is auto-detected by `Qt Creator` or add one manually.
 
+It may be necessary to add:
+
+```
+set(CMAKE_PREFIX_PATH "<path_to_Qt>/5.9.1/gcc_64/lib/cmake")
+```
+
+to the root `CMakeLists.txt` file in `react-native-desktop` repo in order for QtCreator to find the necessary packages.
+
 ### Build react-native-desktop with Qt Creator
 
 In `Qt Creator` select menu `File`->`Open File or Project...`.  
--- a/Development-linux.md
+++ b/Development-linux.md
@@ -33,7 +33,7 @@ Make sure that `CMake` tool is auto-detected by `Qt Creator` or add one manually
 It may be necessary to add:
 
 ```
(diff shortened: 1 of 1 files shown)


Development-linux.md
```diff
@@ -30,6 +30,14 @@ Install:
 Launch `Qt Creator`, select `Tools`->`Options..`->`Build & Run`->`CMake` tab.
 Make sure that `CMake` tool is auto-detected by `Qt Creator` or add one manually.
 
+It may be necessary to add:
+
+```
+set(CMAKE_PREFIX_PATH "<path_to_Qt>/5.9.1/gcc_64/lib/cmake")
```
vkjr: @rcullito, I have `export PATH=$PATH:/Users/vkjr/Qt/5.11.2/clang_64/bin/` set (of course for your exact path) in `.bash_profile` and it is enough on mac to avoid setting CMAKE_PREFIX_PATH. Did you try that?

rcullito: thanks for responding to this @vkjr. Yes, I've always had: `/home/robert/Qt/5.9.1/gcc_64/bin` in my path, including when I first triend compiling with QtCreator yesterday. For what it's worth I also have the same path set inline above as an env var under CMAKE_PREFIX_PATH. So on linux mint, at least, I found this was necessary to explicitly set it in the CMakeLists.txt file. Although I can't say if this would affect all linux users equally. cc @MaxRis

rcullito: small update: also found this to be the case on linux after updating to 5.11.2 also

MaxRis: Time to time it makes sense to delete manually CMakeListsCache.txt and others cmake cache files from the project left from previous builds.

rcullito: ok, thanks also then @MaxRis. I too am on QtCreator 4.7.1 and upgraded my `cmake` just now from 3.5.1 to 3.13.0-rc3 and the issue is still persisting. I'll close this PR down though and remove it if this has not been a problem for others during install as I don't want to overcomplicate the documentation.

PR #394, opened 2018-11-13
Fixed inverted FlatList behavior on resize

### LINKED ISSUE #391: inverted ListView resizes in the wrong direction
## Description

When a ListView with prop `inverted` `true` is part of a parent view and one of the childs changes in size, the ListView shrinks in the wrong direction. The result is that the content
of the ListView overlaps with the growing view and on the other side there is an empty space

screenshot from 2018-11-09 15-44-47

Probably related the scroll direction isn't the right either.

vkjr:
Fixes #391

Scale transformation had to be reapplied after `FlatList` changed size. Now when it is done `FlatList` resizes correctly.


vkjr:
@yenda, this change just merged to `react-native-desktop`. You can create a new tag pointing to this commit and use it in your blocked branch - https://github.com/status-im/status-react/pull/6652
This way you will be able to test it and merge these changes along with yours.

But I just found one moment - looks like @PombeirP switched `status-react` to tag `v0.56.1_2` that doesn't point to `react-native-desktop` master. 
@PombeirP, could you please elaborate on this? Do we have something in `master` branch that breaks `status-react` and fixed in your branch? If so, maybe we should commit your branch to master? Otherwise we are blocked from moving `status-react` to newer tags in `react-native-desktop`.

PombeirP:
@vkjr yeah I was told I could merge the Windows commit to master because in any case the package was locked, but then someone created the `v0.56.1_1` tag which included my commit and therefore status-react build was broken. So I went ahead and created `v0.56.1_2` on another branch that reverted my commit from master so that we wouldn't need to revert on `master` (that had happened already once in the past).

I'm hoping to merge the [Windows PR](https://github.com/status-im/status-react/pull/6465) soon (waiting for testing) so this recurrent problem will be gone.

yenda:
@vkjr the PR breaks the app but I don't get any error in the logs to explain it

here's what I get when I go to profile advanced settings https://imgur.com/a/qZoGr4O

clicking home doesn't work
there is no mailservers shown
clicking edit asks me if I want to use a particular mailserver
it's kind of messed up

vkjr:
@yenda, you are right, my bad. Wanted to fix inverted lists and break non-inverted...
Commited fix - https://github.com/status-im/react-native-desktop/commit/be5617f0824265da4573049627df47b8bf9ffdd5

vkjr:
@PombeirP, if your changes break `status-react`, maybe it worth removing them from `react-native-desktop` master branch temporarily? To not block switching to new tags? And when your PR fixed we can get them back.

PombeirP:
Sure, we can temporarily revert master. The Windows PR will probably be merged later today, but if this is blocking people, let's do it.

vkjr:
@PombeirP, thanks!

PR #395, opened 2018-11-14
Fixed inverted FlatList behavior on resize (2nd attempt)

### LINKED ISSUE #391: inverted ListView resizes in the wrong direction

vkjr:
Fixes #391 

Scale transformation had to be reapplied after FlatList changed size. Now when it is done FlatList resizes correctly.

@yenda, I don't what is going on, but all the changes from previous PR merge doesn't show up in `master`... So I'm creating a new PR with all FlatList fixes together.


maxhora (approved):

vitvly (approved):

yenda (approved):

vkjr:
@siphiuel, @MaxRis, do you see commits from this PR in `master`?
I believe something wrong happens with github. I do not see them...

MaxRis:
screen shot 2018-11-14 at 6 48 29 pm
At the moment they are discoverable via browser client. But the hashes of commits are different?

vkjr:
There was a discussion in #status-desktop channel on topic. It appeared to be an incorrect force push that overwrote master.

PR #412, opened 2018-11-20
Fixed modal closing

### LINKED ISSUE #326: Cannot close modal window
The problem is reproducible using the code snippet below:
```
import React, {Component} from 'react';
import { AppRegistry, Modal, Text, View, TouchableHighlight } from 'react-native';

class rn_popup extends Component {
  state = {
    modalVisible: true,
  };

  setModalVisible(visible) {
    this.setState({modalVisible: visible});
  }
    render () {
          return (
                  <View>
                  <View style={{backgroundColor: 'rgba(0,0,0,0.4)'}}>
                    <Text>some text here</Text>
                    <Text>and more text here</Text>
                  <TouchableHighlight style={{backgroundColor: '#00ff22', opacity: 1.0}} 
            onPress={() => {this.setModalVisible(true);}}>
                  <View >
                    <Text>some text</Text>
                  </View>
                  </TouchableHighlight>
                  </View>
                  <Modal transparent={true} visible={this.state.modalVisible}>
                  <View style={{flex: 1,
                                flexDirection: 'column',
                                justifyContent: 'center',
                                borderRadius:     20,
                                position: 'absolute',
                                top:        20,
                                left: 80,
				backgroundColor: 'blue',
                                width: 200,
                                height: 200,
                                alignItems: 'center'}}>
                  <TouchableHighlight onPress={() => {this.setModalVisible(!this.state.modalVisible);}}>
                    <Text>modal text</Text>
                  </TouchableHighlight>
                  </View>
                  </Modal>
                  </View>
          );
        }
}

AppRegistry.registerComponent('rn_popup', () => rn_popup);
```

Clicking on `modal text` inside the modal should close it, but it has no effect.

vkjr:
Fixes #326

Now modal window can be closed.

Usually `RootView` intercepts all mouse events and sends them to js side. But qml Popup creates not as a RootView child, but as a child of special `Overlay`. I've added new qml item - MouseEventsInterceptor that catches mouse events and sends them to js, like RootView. When it is added as a Popup `contentItem` modal works as supposed.


ReactQt/runtime/src/reactplugin.cpp
```diff
@@ -14,6 +14,13 @@
 #include "reactplugin.h"
 #include "utilities.h"
 
+#include <QLoggingCategory>
+
 void ReactPlugin::registerTypes(const char* uri) {
     utilities::registerReactTypes();
+    QLoggingCategory::setFilterRules("UIManager=true\n"
```
MaxRis: Should it be removed before merge?

vkjr: I found it to be quite convenient when you can change logger settings in code instead of setting env var.
Maybe we can leave it commented for debug purpose, wdyt?

MaxRis: sounds good, or maybe under #ifdef _DEBUG condition

vitvly (approved):

maxhora (approved):

Issue #420, opened 2018-11-23
qCDebug logging blocks UI updates on Windows


madhavarshney:
## Environment
Windows 10
React Native Desktop 0.56
Qt 5.11.2

## Description
When running a react-native-desktop app on Windows, there is a long delay between when the app window opens and the UI is painted, and when clicking the buttons/options in the list in RNTester.  This issue also causes some animations to not work on Windows.  (The animation issue was discovered by running the old Facebook 2048 game example with react-native-desktop.)  Based on my testing, this is caused by large chunks of text being printed to the console with `qCDebug`, blocking UI updates.

## Potential Fix
Disable `qCDebug` calls in `viewmanager.cpp`, `flexbox.cpp`, `uimanager.cpp`, and `utilities.cpp` or add a flag to enable `qCDebug` calls and disable them by default. (Calls in `websocketmodule.cpp` and `networking.cpp` can also be disabled, although they did not cause much of a slowdown.)  IMO, these debug logs are unnecessary for most people.

## Steps to Reproduce
Create a new app on Windows and change `AppRegistry.registerComponent('APP_NAME', () => APP_NAME);` to `AppRegistry.registerComponent('APP_NAME', () => require('RNTesterApp'));`, and run the app.

## Expected Behavior
A snappy UI.

## Actual Behavior
Long delays, broken animation.

PR #425, opened 2018-11-25
Don't force RemoteServerConnection and don't launch ubuntu-server.js unnecessarily

### LINKED ISSUE #424: Ubuntu-server.js does not need to be run separately
## Environment
Windows 10
Qt 5.11.2
React Native Desktop 0.56

## Description
React-native-desktop starts its own process of `ubuntu-server.js` and connects to that instead of the separate server started by `react-native run-desktop`.

https://github.com/status-im/react-native-desktop/blob/e4dbb4a7e6c8bbd911193904a675b3d665c5a03f/ReactQt/runtime/src/communication/serverconnection.cpp#L58-L67

https://github.com/status-im/react-native-desktop/blob/e4dbb4a7e6c8bbd911193904a675b3d665c5a03f/local-cli/desktop/runDesktop.js#L100-L106

This only happens on Windows, however, as react-native run-desktop forces the `RemoteServerExecutor` regardless of what is set in `main.cpp`, but `run-app.bat` does not use the argument `--executor` passed by `runDesktop.js` (this should be a little more clear after looking at the code).

https://github.com/status-im/react-native-desktop/blob/e4dbb4a7e6c8bbd911193904a675b3d665c5a03f/local-cli/desktop/runDesktop.js#L74-L78

https://github.com/status-im/react-native-desktop/blob/e4dbb4a7e6c8bbd911193904a675b3d665c5a03f/local-cli/generator-desktop/templates/run-app.bat.in#L9-L10

https://github.com/status-im/react-native-desktop/blob/e4dbb4a7e6c8bbd911193904a675b3d665c5a03f/local-cli/generator-desktop/templates/run-app.sh.in#L29-L36

The solution to the problem I have just described is to remove the part of code that forces `RemoteServerExecutor`.  This would also mean that `ubuntu-server.js` would only be needed to be run separately in special circumstances (if I am correct) where `RemoteServerExecutor` is needed instead of `LocalServerExecutor`.

madhavarshney:
## Motivation
Fixes #424.

### Test Plan
N/A

### Related PRs
N/A

### Release Notes
N/A

### DIFF
--- a/docs/CreateNewApp.md
+++ b/docs/CreateNewApp.md
@@ -28,15 +28,12 @@ react-native desktop
 ```
 
 ### Run the project
-If you're using macOS, run these commands in 2 different shells (from `DesktopSampleApp` dir):
+If you're using macOS, run this command in a different shell (from `DesktopSampleApp` dir):
 ```sh
 npm start #starts bundler
 ```
-```sh
-node node_modules/react-native/ubuntu-server.js #starts js server
-```
 
-Afterwards, in a 3rd shell execute:
+Afterwards, execute in a another shell:
 ```sh
 react-native run-desktop
 ```
@@ -44,7 +41,7 @@ Compilation of desktop project will start. When it finished and app run you can 
 
 ![](./media/react-native-desktop-new-app.png)
 
-**Note:** On non-macOS systems, `npm start` and `node ubuntu-server.js` will be executed automatically by the above command
+**Note:** On non-macOS systems, `npm start` will be executed automatically by the above command
 
 
-**If you want to know why you had to start bundler and js server** - check the doc on [how react-native-desktop works internally](docs/HowRNDesktopAppWorks.md)
+**If you want to know why you had to start bundler** - check the doc on [how react-native-desktop works internally](docs/HowRNDesktopAppWorks.md)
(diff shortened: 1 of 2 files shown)


MaxRis:
initially, there are 2 ways to start `ubuntu_server.js` - as separate `node ubuntu_server.js` process which communicates with native app via socket or as separate `node ubuntu_server.js --pipe` process which communicates via process standard input/output.
For development it's more convenient to use communication via sockets ( js output goes to separate console ), for production - via standard process input/output (no dependency on socket availability; probably, implementation complexity is less).
Current supposed logic is to use sockets on `react-native run-desktop` by passing additional command line param to overwrite default behavior to expect ubuntu_server launched to communicate via standard process input/output.

madhavarshney:
@MaxRis - Makes sense.  However, I think it's much more convenient to use the `--pipe` option for development as well for two reasons: there isn't an extra window to manage and you don't have to manually launch the server on MacOS.  I think the trade off, which is outputting the JS console output to the same terminal window as `react-native run-desktop` is worth it, as for debugging JS code, devtools is usually used.  We can also minimize the amount of console output as I said in #420, or color code the output.

madhavarshney:
Closing in favor of #436.

PR #426, opened 2018-11-27
Fix NetInfo module

### LINKED ISSUE #378: Remove dependency on hard-coded URL in NetInfo module
### Description
Right now, network availability relies on www.google.com to be accessible. It is necessary to devise a way to specify a pool of URLs to use for internet availability checking.

vkjr:
Fixes #378

- `NetInfo` module updated to correspond [documentation](https://facebook.github.io/react-native/docs/netinfo). Deprecated calls not supported.
- RNTester NetInfo example updated: deprecated `change` event substituted with `connectionChange`
- `NetInfo.setConnectionCheckUrl` added for substituting default google pinging with any alternative

### DIFF
--- a/Libraries/Network/NetInfo.desktop.js
+++ b/Libraries/Network/NetInfo.desktop.js
@@ -1,0 +1,254 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+const Map = require('Map');
+const NativeEventEmitter = require('NativeEventEmitter');
+const NativeModules = require('NativeModules');
+const Platform = require('Platform');
+const RCTNetInfo = NativeModules.NetInfo;
+
+const NetInfoEventEmitter = new NativeEventEmitter(RCTNetInfo);
+
+const DEVICE_CONNECTIVITY_EVENT = 'networkStatusDidChange';
+
+type ChangeEventName = $Enum<{
+  connectionChange: string,
+  change: string,
+}>;
+
+type ReachabilityStateIOS = $Enum<{
+  cell: string,
+  none: string,
+  unknown: string,
+  wifi: string,
+}>;
+
+type ConnectivityStateAndroid = $Enum<{
+  NONE: string,
+  MOBILE: string,
+  WIFI: string,
+  MOBILE_MMS: string,
+  MOBILE_SUPL: string,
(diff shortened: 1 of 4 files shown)


vkjr:
Just found that this fix doesn't work well when wifi connnected but there is no network access :(

MaxRis:
@vkjr it might make sense to combine your changes with configurable URL check. As much as possible testing here, probably, plays significant role in understanding efficiency of final implementation (how it works as part of status-react also)

vkjr:
@maxris, agree, I'm also now thinking that we won't be able to check real connectivity without pinging some URLs.

vkjr:
Pushed fixes. Now we again checking connectivity by pinging google. But connection type determined from available interfaces (wifi/ethernet). Also there is a `NetInfo.setConnectionCheckUrl` that allows user to use another host to ping.

vitvly (approved):
Awesome! One thing that I'm wondering though is if it would be possible to modify setConnectionCheckUrl() to accept an array of URLs to check (if just one is accessible, then we're fine)

vkjr:
@siphiuel, I'm also was considering this option. Here what stopped me:
When some app want to check connectivity, it usually knows what exact resources it want to check. If we allow providing a big list of resources to be checked internally - that would increase waiting timeout. We wait 3 seconds for one resource. For 3 resources that would be already max 9 seconds, quite a long time.
Instead any app that has particular knowledge about resources it want to check, can do this explicitly in the app. It can set urls one by one and check them. Also it can then display to user what resource it checks right now. That will give an app chance to show meaningful info to user who waits.
wdyt?

siphiuel:
@vkjr agree, good point, especially about transparency to the app

maxhora (approved):

MaxRis:
I think that accepting of pool of URLs can be implemented with async triggering of limited number of URLs from the list. For example, react-native-desktop can send requests to randomly selected 3 URLs from the pool; once some URL returns OK response, it aborts others not finished requests. Other logic is similar to what we have currently.

vkjr:
@MaxRis, ok, added this functionality as a separate feature #430.

MaxRis:
@vkjr thanks!

oskarth:
What's going on with this PR? It appears to be a blocker for desktop release.

vkjr:
@oskarth, this PR fails on testing and was put aside to work on other release-related one.
Actually it is not blocker to status desktop release and not even related to it. We had a discussion and created another issue for getting rid of `google.com` connection - https://github.com/status-im/status-react/issues/6961

This one should be removed from release task list. Will do it now.

vkjr:
Fixed failing build. Merging this one.

PR #436, opened 2018-12-15
[WORK IN PROGRESS] JS Executor based on Qt WebEngine

maxhora:
Removes dependency on ubuntu_server


madhavarshney:
Hey @MaxRis - I actually worked on a similar PR with the executor based on QJSEngine instead of QWebEngine, which also fixed remote JS debugging.  I'm not sure why you chose QWebEngine over QJSEngine as it adds another required Qt dependency and runs a full Chromium browser. (Isn't Electron better then?)  I was able to get almost everything working.

MaxRis:
Hey @madhavarshney , seems, you are right, QJSEngine looks much more suitable for the current needs. I might need to update my PR to use QJSEngine instead.

PR #438, opened 2018-12-24
Added default user-agent to get requests

vkjr:
Qt has an issue with retrieving data from some servers because it doesn't add `user-agent` field to get requests.
This PR fixes that.


vitvly (approved):
Cool!

Issue #345, opened 2018-09-24
`View`: inline styles are not applied


rasom:
Inline style props are ignored. Here example with `backgroundColor` in desktop version:

![](https://user-images.githubusercontent.com/2364994/45942509-6525dc80-bfeb-11e8-852e-9833f75505d8.png)

code:

```js
import React, { Component } from 'react';
import {
AppRegistry,
StyleSheet,
Text,
View
} from 'react-native';


class App extends React.Component {
render() {
return (
<View style={styles.container}>
<View backgroundColor={'green'}>
<Text>Open up App.js to start working on your app!</Text>
</View>
</View>
);
}
}

const styles = StyleSheet.create({
container: {
flex: 1,
backgroundColor: '#fff',
alignItems: 'center',
justifyContent: 'center',
},
});

AppRegistry.registerComponent('DesktopSampleApp', () => App);
```

mobile version:
![](https://user-images.githubusercontent.com/2364994/45942502-5d663800-bfeb-11e8-96e1-73c3f88d2397.png)

code:

```js
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';

export default class App extends React.Component {
render() {
return (
<View style={styles.container}>
<View backgroundColor={'green'}>
<Text>Open up App.js to start working on your app!</Text>
</View>
</View>
);
}
}

const styles = StyleSheet.create({
container: {
flex: 1,
backgroundColor: '#fff',
alignItems: 'center',
justifyContent: 'center',
},
});
```

PR #451, opened 2019-02-07
Changes for running status mobile ui

vkjr:
- getInitialUrl() added to Linking module
- #345 fixed

### DIFF
--- a/ReactQt/runtime/src/linkingmanager.cpp
+++ b/ReactQt/runtime/src/linkingmanager.cpp
@@ -38,7 +38,7 @@ void LinkingManager::getInitialURL(const ModuleInterface::ListArgumentBlock& res
                                    const ModuleInterface::ListArgumentBlock& reject) {
     Q_D(LinkingManager);
     Q_UNUSED(reject);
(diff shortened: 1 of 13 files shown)


ReactQt/runtime/src/linkingmanager.cpp
```diff
@@ -34,6 +34,13 @@ void LinkingManager::canOpenURL(const QString& url,
     resolve(d->bridge, QVariantList{true});
 }
 
+void LinkingManager::getInitialURL(const ModuleInterface::ListArgumentBlock& resolve,
+                                   const ModuleInterface::ListArgumentBlock& reject) {
+    Q_D(LinkingManager);
+    Q_UNUSED(reject);
+    resolve(d->bridge, QVariantList{QString()});
```
madhavarshney: Should return null instead of an empty string based on [the documentation](https://facebook.github.io/react-native/docs/linking#getinitialurl).

vkjr: @madhavarshney, good point, thanks!

PR #458, opened 2019-03-01
Yeoman version bumped

vkjr:
Yeoman version updated because its dependencies has vulnerabilites

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -838,16 +838,25 @@
         }
       }
     },
+    "@mrmlnc/readdir-enhanced": {
+      "version": "2.2.1",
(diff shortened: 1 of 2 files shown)


mandrigin:
Thanks! Appreciate!

madhavarshney:
This probably breaks react-native-desktop project creation.  Yeoman generator code needs to be updated.

vkjr:
Hm, strange... I tried `react-native desktop` command with these changes before merging PR and it worked fine. But now i double-checked and it breaks...
@madhavarshney, thanks for alertness, will fix this.

PR #467, opened 2019-03-17
[FIX] hostname -i with lowerecase

hlolli:
On linux with GNU ineturils 1.9.4 (current stable release) the use of an uppercase hostname info causes error

```
➜  ~ hostname -I                                        ~
hostname: invalid option -- 'I'
Try 'hostname --help' or 'hostname --usage' for more information.
```

but lowercase works

```
➜  ~ hostname -i                                        ~
127.0.1.1 
```

Not sure if this cross platform command, I'd only assume so that it works on MacOs too, if not, maybe there needs to be a boolean there before.

### DIFF
--- a/local-cli/generator-desktop/templates/run-app.sh.in
+++ b/local-cli/generator-desktop/templates/run-app.sh.in
@@ -13,7 +13,7 @@ plugins_path=""
 asset_path="share"
 executor=""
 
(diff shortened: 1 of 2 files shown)


vkjr:
Fixed during project restructuring.

PR #488, opened 2019-08-07
InstallPrerequisites: Add Archlinux/Manjaro

zaoqi:


vkjr:
Thanks for the pr. 
For now Ubuntu will be the only supported linux OS, so docs for other OSes won't be included.

